PR #186, opened 2021-07-13
Update Arch Linux package to official one

metalmatze:
CUE has become an official Arch Linux package:

https://twitter.com/Sh1bumi/status/1413892649984036866
https://archlinux.org/packages/community/x86_64/cue/


myitcv:
@metalmatze - thanks so much for doing this. And apologies for not pinging you about the migration - that slipped my mind.

PR #196, opened 2021-07-23
deps: use latest version of Hugo

myitcv:
This addresses advisory:

https://github.com/advisories/GHSA-8j34-9876-pvfq


cedricgc:
Quick skim through the site and everything looks as I expected

PR #198, opened 2021-08-03
update Arch Linux cue package

cbrake:
I think the Arch Linux package is now called cue:

https://archlinux.org/packages/community/x86_64/cue/

### DIFF
--- a/content/en/docs/install/_index.md
+++ b/content/en/docs/install/_index.md
@@ -29,7 +29,7 @@ brew install cue-lang/tap/cue
 Arch Linux has an official package that you can install by running:
 
 ```
-pacman -S cuelang-bin
+pacman -S cue
 ```
 
 ## Install CUE from source


myitcv:
@metalmatze - you've been involved in previous PRs, please can you provide some guidance here?

Would also be good to have a conversation about how we support and document third-party installation sources, along the lines of the conversation we are having for MacPorts. They are a really important source of new users, but important to get the process right.

metalmatze:
That's correct. CUE is now an official Arch Linux package. That is why it's called cue now and one can use pacman directly rather than some AUR helper.
LGTM

myitcv:
@metalmatze - thanks for confirming.

@cbrake. Apologies for the delay here. Please can you assert the DCO then we can get this merged?

https://github.com/cue-lang/cue/blob/master/doc/contribute.md#asserting-a-developer-certificate-of-origin

PR #203, opened 2021-11-14
content: add import statement to protobuf docs

dylankb:
This adds an import statement to the Protobuf section of the Cue
documentation site to help disambiguate the correct package to import. See
related discussion https://github.com/cue-lang/cue/discussions/1362.

Signed-off-by: Dylan Barnard <dylan.barnard@gmail.com>

### DIFF
--- a/content/en/docs/integrations/protobuf.md
+++ b/content/en/docs/integrations/protobuf.md
@@ -109,21 +109,24 @@ resides in
 can be converted to CUE using the following Go code
 
 {{< highlight go >}}
-  import (
-  	"cuelang.org/go/cue/format"
-  	"cuelang.org/go/encoding/protobuf"
-  )
+import (
+	"cuelang.org/go/cue/format"
+	"cuelang.org/go/encoding/protobuf"
+	// ...
+)
 
-file, err := protobuf.Extract("basic.proto", nil, &protobuf.Config{
-    Paths: []string{ /* paths to proto includes */ }],
-})
+func main() {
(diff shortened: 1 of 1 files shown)


content/en/docs/integrations/protobuf.md
```diff
@@ -109,6 +109,11 @@ resides in
 can be converted to CUE using the following Go code
 
 {{< highlight go >}}
+  import (
+  	"cuelang.org/go/cue/format"
+  	"cuelang.org/go/encoding/protobuf"
+  )
+
```
dylankb, line 116: Following pattern of highlighting ["notable" cuelang.org packages](https://github.com/cuelang/cuelang.org/blob/42bcb4ee6fb146b64d98005a3f77c66e0ae8cb71/content/en/docs/integrations/openapi.md?plain=1#L27-L28) in the given code snippet rather than providing a strictly complete listing of all import statement used in the program.

myitcv: @dylankb apologies for the delay in replying here.

The problem however is that the code snipper that follows is not valid outside of a function. So I think we need some additional context (i.e. wrap in a simple `func main`) for what follows possibly with `// ...` to indicate this is an incomplete snippet.

dylankb: @myitcv Makes sense, made the suggested changes. Let me know what you think.

myitcv (approved):
LGTM thanks. There's one thing that I will fix up but given the length of time since my last review (apologies) I will do that in a subsequent PR.

PR #225, opened 2022-02-10
doc: fix up install instructions

myitcv:
(see commit message)


myitcv:
Preview the result here: https://deploy-preview-225--cue.netlify.app/docs/install/

Note the redirect will only on the live site (which is frustrating)

PR #229, opened 2022-02-22
docs/install: fix arch install instructions

myitcv:
(see commit message)

With thanks to @cbrake who submitted the original fix in #198. That PR
Closes #198.


metalmatze:
Uh nice! :tada:

PR #237, opened 2022-03-09
Fix broken anchor links

deining:
This PR fixes two broken anchor links.


myitcv:
Thanks for these fixes @deining and apologies we didn't get to them in a timely way.

I have picked them up in https://review.gerrithub.io/c/cue-lang/cuelang.org/+/1194116 now that we have moved to the new https://cuelang.org site.

PR #240, opened 2022-03-23
Fix Value.Decode() error handling example

paulhammond:
This PR updates the error handling example in "Copy CUE values into Go values" to work as documented. See https://github.com/cue-lang/cue/issues/1592 for additional context.

Note: this does not update the example to use `cue.Context`. As a newcomer to cue I'm not confident enough to make that change.

### DIFF
--- a/content/en/docs/integrations/go.md
+++ b/content/en/docs/integrations/go.md
@@ -170,12 +170,12 @@ _ = i.Value().Decode(&x)
 fmt.Println(x)
 
 i, _ = r.Compile("test", `{B: "foo"}`)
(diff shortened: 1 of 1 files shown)


myitcv:
@paulhammond - apologies for the terrible response rate here. Since you raised this PR we have upgraded https://cuelang.org such that all the code examples are fully tested at build time. 

The page you sought to fix has been updated on https://cuelang.org/docs/concept/how-cue-works-with-go/

paulhammond:
Thanks for the update - the wider fix you've made sounds like a much better solution than this standalone PR!

PR #247, opened 2022-04-27
play: update to latest playground version

rogpeppe:
(see commit message)


myitcv (approved):
LGTM.

Just noting here that my example didn't work:

```cue
x: 5
x: 4

y: 1
y: 2
```

But we can follow-up on that separately.

myitcv (approved):
Looks even better to me now!

PR #238, opened 2022-03-13
doc: translate  /about pages

h1z3y3:
Add Chinese translation of /about

PR #250, opened 2022-06-16
theme: use docsy as hugo module

h1z3y3:
As docsy has supported be used as hugo module, we can use it as hugo module too.


myitcv:
@h1z3y3 thanks for the change here. Per my comment in https://github.com/cue-lang/cuelang.org/pull/238#issuecomment-2088251795, we have since upgraded https://cuelang.org to a new look site that does not use docsy. So this PR has since been superseded.

PR #259, opened 2022-07-18
gerritstatusupdater: set success status in nil branch of defer

myitcv:
This correctly captures the cases where we successfully perform an early
return from our handler.

As part of this change, don't pother write "pong" in response to a ping.
A 200 response code is sufficient.


myitcv:
> commit message nit: pother write -> bother writing

Done.

PR #261, opened 2022-07-21
gerritstatusupdater: fail any completed job that did not succeed

myitcv:
(see commit message)


mvdan (approved):
Note that this might include manually cancelled jobs.

myitcv:
> Note that this might include manually cancelled jobs.

Indeed. But I still think we need to consider these as failed, because by definition they did not succeed. It's precisely the case of manually cancelled jobs that currently leaves Gerrit CLs in a limbo state. I think the best we can do is mark them as failed; we definitely can't mark them as having passed.

PR #265, opened 2022-07-30
netlify.toml: add redirect rule for Gerrit CLs

### LINKED ISSUE #264: website: redirect cuelang.org/cl like we do for cuelang.org/issue
URLs like https://cuelang.org/issue/123 redirect to our issue tracker, currently https://github.com/cue-lang/cue/issues. It helps to use URLs like that because, if we ever need to move the issue tracker, the URLs can be redirected accordingly.

We should do the same for CLs; for example, https://cuelang.org/cl/538063 is just a 404, when it could redirect to https://review.gerrithub.io/c/cue-lang/cue/+/538063. As an extra bonus, note how much longer the real Gerrit URL is, so that will also help for commit messages.

cedricgc:
Currently we redirect cuelang.org/issue/123 to our issue tracker.
This change will add a similar rule for GerritHub CLs. For example,
https://cuelang.org/cl/538063 will now redirect to
https://review.gerrithub.io/c/cue-lang/cue/+/538063 for future
portability and convenience.

Fixes #264

Signed-off-by: Cedric Charly <cedric.charly@gmail.com>

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -13,6 +13,12 @@ HUGO_VERSION = "0.89.4"
 command = "bash build.bash -b $DEPLOY_URL"
 
 [[redirects]]
+  from = "/cl/*"
+  to = "https://review.gerrithub.io/c/cue-lang/cue/+/:splat"
+  status = 302
+  force = true
+
+[[redirects]]
   from = "https://cuelang.org/issue/*"
   to = "https://github.com/cue-lang/cue/issues/:splat"
   status = 302


cedricgc:
I used the partial path for the redirect. If the redirect works on the deploy preview I would be happy to update the other rules to use the same format.

myitcv (approved):
LGMT pending preview review

myitcv:
Appears to work for me: https://deploy-preview-265--cue.netlify.app/cl/541788

PR #270, opened 2022-09-08
Update logic.md

ryan-ju:
Fixed grammar


myitcv:
Thanks for the contribution, @ryan-ju!

It is indeed a bit of a clumsy sentence! However, I'm not actually clear your changes helps to make the sentence any clearer.

> This page explains the core concept on which pretty much everything that CUE depends

In fact I think it actually reads incorrectly as a result. 

Better would be something like:

> This page explains the core concepts that underpin CUE.

Plain and simple language.

That said, the docs are going to get an overhaul so we will revisit this later. 

Thanks

PR #271, opened 2022-09-19
internal/functions: be more careful about gerrit attention sets

mvdan:
(see commit message)


mvdan:
No way to really test this other than pushing to master and testing it out, I think.

myitcv (approved):
LGTM

PR #281, opened 2022-11-30
all: add short link to community calendar

myitcv:
See commit message.


myitcv:
Only question here is whether we should use the top level namespace, or use `/s/community-calendar` to leave the top level less crowded. I'm minded towards "yes, use the `/s` namespace".

4ad:
`/s` sounds good.

4ad (approved):

PR #285, opened 2022-12-19
internal/ci: move to using CUE as a source of truth for CI

myitcv:
(see commit messages)


.github/workflows/trybot.yml
```diff
@@ -1,61 +1,53 @@
...
-      with:
-        node-version: '18.9.0'
-    - name: Install Go
-      uses: actions/setup-go@v3
-      with:
-        go-version: '1.19.1'
-    - name: Ensure latest CUE
-      # The latest git clean check ensures that this call is effectively
-      # side effect-free. Using GOPROXY=direct ensures we don't accidentally
-      # hit a stale cache in the proxy.
```
mvdan, line 28: are you losing comments on purpose?

.github/workflows/trybot.yml
```diff
@@ -38,6 +43,9 @@ jobs:
         working-directory: ./play
       - name: Test
         run: go test ./...
+      - name: Test play
+        run: go test ./...
+        working-directory: ./play
```
mvdan, line 49: I notice that this is added. Just checking that this is on purpose, and I wonder if it was harmful that it wasn't there before.

internal/ci/github/trybot.cue
```diff
@@ -31,19 +31,22 @@ trybot: _base.#bashWorkflow & {
...
 	on: {
 		push: {
-			branches: [_#defaultBranch] // do not run PR branches
-		}
-		repository_dispatch: {}
-		pull_request: {
-			branches: ["**"]
+			branches: ["trybot/*/*", _#defaultBranch, _base.#testDefaultBranch] // do not run PR branches
 		}
+		pull_request: {}
```
mvdan, line 36: I think we can omit this?

internal/ci/netlify/netlify.cue
```diff
@@ -0,0 +1,117 @@
...
+
+		{{- range .redirects}}
+		[[redirects]]
+		  from = {{printf "%q" .from}}
+		  to = {{printf "%q" .to}}
+		  status = {{printf "%d" .status}}
+		  force = {{printf "%v" .force}}
+		{{end}}
+		"""
+	template.Execute(tmpl, #input)
```
mvdan, line 118: I think this warrants a TODO - this TOML "encoding" with a template is a bit icky :)

internal/ci/netlify/netlify.cue
```diff
@@ -0,0 +1,117 @@
...
+		[context.deploy-preview]
+		command = {{printf "%q" .command}}
+		{{end}}
+
+		{{- range .redirects}}
+		[[redirects]]
+		  from = {{printf "%q" .from}}
+		  to = {{printf "%q" .to}}
+		  status = {{printf "%d" .status}}
+		  force = {{printf "%v" .force}}
```
mvdan, line 115: nit: since you're using %d and %q for ints and strings, use %t for booleans

mvdan (approved):

.github/workflows/trybot.yml
mvdan: I guess you are, thinking about it again, since we'll have the comments in the CUE source of truth.

internal/ci/github/trybot.cue
myitcv: Why? We need and want the trybot workflow to run for pull requests.

internal/ci/netlify/netlify.cue
myitcv: Done.

internal/ci/netlify/netlify.cue
myitcv: Done.

.github/workflows/trybot.yml
myitcv: There are no tests in play, there might be some in the future. Adding this really to reduce the diff. There is zero impact because there are no tests.

.github/workflows/trybot.yml
myitcv: Yep, exactly.

internal/ci/github/trybot.cue
mvdan: > By default, a workflow only runs when a pull_request event's activity type is opened, synchronize, or reopened. To trigger workflows by different activity types, use the types keyword.

i.e. we only need to set this object, with some of its fields, if we want something other than the default behavior.

myitcv: Resolving per offline conversation.

PR #300, opened 2023-02-14
Netlify: Fix baseUrl for preview-deploy

JorindeUsMedia:
Issue: Base url was not set correctly on preview builds which causes e.g. svg sprites to be blocked because of url mismatches Based on answers here: https://discourse.gohugo.io/t/correct-baseurl-on-netlify-previews/10429 we need to use $DEPLOY_PRIME_URL instead of $DEPLOY_URL

Signed-off-by: Jorinde Reijnierse <jorinde.reijnierse@usmedia.nl>

### DIFF
--- a/internal/ci/netlify/netlify.cue
+++ b/internal/ci/netlify/netlify.cue
@@ -60,7 +60,7 @@ config: #config & {
 		}
 	}
 
-	context: "deploy-preview": command: "bash build.bash -b $DEPLOY_URL"
+	context: "deploy-preview": command: "bash build.bash -b $DEPLOY_PRIME_URL"
 
 	redirects: [...{force: true, status: 302}]
 	redirects: [{
--- a/netlify.toml
+++ b/netlify.toml
@@ -12,7 +12,7 @@ HUGO_VERSION = "0.108.0"
 NODE_VERSION = "18.12.1"
 
 [context.deploy-preview]
-command = "bash build.bash -b $DEPLOY_URL"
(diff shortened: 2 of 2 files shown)


myitcv:
Thanks @JorindeUsMedia. What change/fix should we observe as a result of making this change?

JorindeUsMedia:
@myitcv Before the icons were not working because the baseUrl was not set correctly which caused a mismatch in urls which caused the svg sprites to be blocked. For instance: https://deploy-preview-299--cue.netlify.app/search/  Here you can see the icons in the navigation are not showing and when you check the Network tab you see the svg's get blocked because their origin url is not matching the current url.  

By changing the baseurl for netlify to $DEPLOY_PRIME_URL this is fixed because it now uses the correct url: https://deploy-preview-300--cue.netlify.app/

Apparantly DEPLOY_PRIME_URL is needed to set the righ url for PR previews in netlify.

myitcv:
@JorindeUsMedia - I've updated this PR to make the change in the source CUE file.

Please can you confirm from your side that the resulting preview is as expected?

JorindeUsMedia:
@myitcv Seems to work as expected :) Icons are working again

myitcv (approved):
LGTM.

myitcv:
Closed by https://github.com/cue-lang/cuelang.org/commit/0a9cd13bb9cfe6fa91435b2ad98ef4c6ab124fbf

PR #310, opened 2023-03-01
Hugo: Add option to set toc-root on a section

JorindeUsMedia:
- Setting params toc_root to true will start the tree table of contents at that section.
- All child sections will use a parent with toc_root set to true as the starting point of the index.
- If current page is not toc_root or doesn't have a parent with toc_root set to true it will fall back on .FirstSection

### DIFF
--- a/hugo/layouts/partials/paging/tree.html
+++ b/hugo/layouts/partials/paging/tree.html
@@ -1,6 +1,25 @@
 <nav class="tree">
-    {{ $navRoot := .FirstSection -}}
     {{ $current := . -}}
+    {{ $navRoot := "" -}}
+
+    {{- if ( or (eq $current .FirstSection) (eq $current.Params.toc_root true)) -}}
+        {{ $navRoot = $current }}
+    {{- end -}}
+
+    {{- if (eq $navRoot "") -}}
+        {{ $parent := .Parent -}}
+        {{- if (eq $parent.Params.toc_root true) -}}
+            {{ $navRoot = $parent -}}
+        {{- else -}}
+            {{ $parent = $parent.Parent -}}
+            {{- if (eq $parent.Params.toc_root true) -}}
+                {{ $navRoot = $parent -}}
+            {{- else -}}
+                {{ $navRoot = .FirstSection -}}
+            {{- end -}}
+        {{- end -}}
+    {{- end -}}
+
     {{ $depth := 1 -}}
     {{ $maxDepth := 3 -}}
 


JorindeUsMedia:
@myitcv I haven't had time to create an example page with some explanation and screenshots of the tree but I'm not sure if that's necessary anymore. Also I wanted to make a recursive function to traverse trough pages parents but I couldn't find a way to do that in a go template so it now has a couple of nested if's checking the parent and the parent of the parent with a max of 3 levels (same as maxdepth of tree).

verdverm:
@JorindeUsMedia I wrote a hugo menu template that walks recursively back up the parent pointers. I use this to highlight and open the current page on load. Maybe it can help. The key is to define a template within the partial an use that for recursion.

https://github.com/hofstadter-io/cuetorials.com/blob/main/layouts/partials/menu-hugo.html

Lyoness:
> Setting params toc_root to true will start the tree table of contents at that section.

Which file(s) would we set `toc_root: true` inside to check this?

myitcv (approved):
LGTM. Tested locally by setting this in `hugo/content/en/docs/language-guide/_index.md`. Will make that change in a follow-up CL as a live example of it in action.

myitcv:
Importing this into Gerrit.

myitcv:
Closed via https://github.com/cue-lang/cuelang.org/commit/17b6e3b7c48d0f182cf178544135a2db95ac612f

JorindeUsMedia:
@verdverm Thanks for the example. We use something similar now which we copied and modified from the Docsy theme. 
The case I had now was that within a content section, for example 'docs' you would have this

Docs
- language-guide
  - language-guide-1
  - language-guide-2 
    - language-guide-2.1 (possibly)
  - language-guide-3 
- how-to
  - how-to-1
  - how-to-2 etc
  - etc..
- tutorials
  - tutorial-1
  - turtorial-2 
  - etc.. 
  
  Before on language-guide-2 page for instance you would see this whole tree in the left-side navigation.
  What we we're trying to do now is to be able to set on a subpage if that page is the toc-root of that section.
  So for instance on language-guide we set toc_root: true
  When you're on a sub page of language guide , for instance on language-guide-2, you only want to see the Tree of the first parent that has toc_root set to true: In this case  language-guide. So on all subpages of language guide you will see this in the left site navigation instead of the whole docs-tree:
  
  - language-guide
  - language-guide-1
  - language-guide-2 
    - language-guide-2.1 (possibly)
  - language-guide-3 
  
  That why I was trying to figure out the navRoot for a page. So starting from for instance language-guide-2: go up the tree and find the first parent thas has toc_root set to true. 
  So this would not be a nested template but a recursive function to get the toc_root parent section of that page.

PR #314, opened 2023-03-06
Styling fixes for docs navigation tree

anne-usmedia:
- Doc nav styling fixes
- Tree styling fixes
- Added 600 weight font
- Breadcrumb styling improvement
- Prev & Next styling improvement

For https://linear.app/usmedia/issue/CUE-160

PR #318, opened 2023-03-08
Hugo: Code block refinements & bugs fixes

JorindeUsMedia:
- Make sure both codeblocks with and without language attribute are rendered correctly. With language attribute hugo renders the highlight shortcode, without it renders a regular pre tag.
- Improve margin of pre tag
- Always keep full padding of pre tag to keep distance to title.
- Make title blue instead of darkgrey to make it stand out more
- Add fallback color for background of highlight because we're leaning on a variable outside the current scope which can cause problems without a default.
- Make highlight title a separate bem-class instead of a bem element inside 'highlight'. This is because the title gets rendered outside the highlight block. In case of absent language attribute it won't even be rendered with the highlight block but with a regular pre.
- Remove old pygments config from toml file. We don't need to set pygments to false because the default of hugo is now Chroma.
- Set lineNumbersInTable to false in config to force hugo to render using modern css instead of tables. Tables give a lot of styling problems within code blocks.
- Change examples lineos=table to lineos=true because we don't want to use tables

For: https://linear.app/usmedia/issue/CUE-151

Test url: https://deploy-preview-318--cue.netlify.app/examples/basic/block/#code-blocks

### DIFF
--- a/hugo/assets/scss/base/base.scss
+++ b/hugo/assets/scss/base/base.scss
@@ -351,7 +351,7 @@ pre {
     border-radius: 5px;
     color: var(--pre-color);
     height: 100%;
-    margin-bottom: 1rem;
+    margin: 0 0 1rem;
     max-width: 100%;
     overflow: auto;
     padding: 1rem 1.25rem;
--- a/hugo/assets/scss/components/highlight-title.scss
+++ b/hugo/assets/scss/components/highlight-title.scss
@@ -1,0 +1,20 @@
+@import '../config/colors';
+@import '../config/typography';
+
+.highlight-title {
+    background-color: var(--pre-background-color, $c-blue--lighter);
+    border-radius: 5px 5px 0 0;
+    color: $c-blue;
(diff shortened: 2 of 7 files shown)


JorindeUsMedia:
I also want to make the highlight-title semibold but I'm waiting on #314 to be merged because that includes the newly added semibold font which we also want to use for the highlight title.

Lyoness (approved):
Great! Importing to Gerrit

Lyoness:
https://review.gerrithub.io/c/cue-lang/cuelang.org/+/550681

JorindeUsMedia:
Rebased and fixed the conflicts @Lyoness @myitcv

myitcv:
Thanks - reimporting now.

Lyoness:
Closes [6a59b72](https://github.com/cue-lang/cuelang.org/commit/6a59b7287edb73f32ce8394479cbe27cffe706d1) as merged.

PR #323, opened 2023-03-09
Hugo: Disable links upcoming pages docs

anne-usmedia:
- Added coming_soon to frontmatter empty pages
- Set page link in nav tree to disabled when coming_soon = true
- Set page link in overview to disabled when coming_soon = true
- Set page link in prev/next to disabled when coming_soon = true
- Added disabled link styling to nav tree links

For https://linear.app/usmedia/issue/CUE-165

### DIFF
--- a/hugo/content/en/docs/howto/constrain-list-ints-no-duplicates/index.md
+++ b/hugo/content/en/docs/howto/constrain-list-ints-no-duplicates/index.md
@@ -1,5 +1,6 @@
 ---
 title: How to constrain a list of integers to have no duplicates
-weight: 
+weight:
 draft: false
+disabled: true
 ---
--- a/hugo/layouts/docs/single.html
+++ b/hugo/layouts/docs/single.html
@@ -10,7 +10,14 @@
                         <h1 class="article__title">{{ .Title }}</h1>
                     </header>
                     <div class="article__content">
-                        {{ .Content }}
+                        {{ if .Params.disabled }}
+                            <div class="todo">
+                                <span class="todo__note">TODO</span>
+                                <span class="todo__desc">{{ T "docs_coming_soon" }}</span>
+                            </div>
+                        {{ else }}
+                            {{ .Content }}
+                        {{ end }}
                     </div>
                     <footer class="article__footer">
                         {{ partial "last-modified.html" . }}
--- a/hugo/assets/scss/components/tree.scss
+++ b/hugo/assets/scss/components/tree.scss
@@ -38,6 +38,16 @@
     &__link {
         display: block;
         text-decoration: none;
+
+        &[disabled] {
+            color: $c-grey--dark;
+            cursor: default;
+            pointer-events: none;
+
+            #{ $self }__text {
+                background-size: 0 0;
+            }
+        }
     }
 
     &__text {
--- a/hugo/content/en/docs/howto/constrain-list-strings-no-duplicates/index.md
+++ b/hugo/content/en/docs/howto/constrain-list-strings-no-duplicates/index.md
@@ -1,5 +1,6 @@
 ---
 title: How to constrain a list of strings with no duplicates
-weight: 
+weight:
 draft: false
+disabled: true
 ---
--- a/hugo/content/en/docs/howto/constrain-list-structs-no-duplicates/index.md
+++ b/hugo/content/en/docs/howto/constrain-list-structs-no-duplicates/index.md
@@ -2,4 +2,5 @@
 title: How to constrain a list of structs with no duplicates
 weight:
 draft: false
+disabled: true
 ---
--- a/hugo/content/en/docs/howto/constrain-max-items-list/index.md
+++ b/hugo/content/en/docs/howto/constrain-max-items-list/index.md
@@ -2,4 +2,5 @@
 title: How to constrain the maximum items in a list
 weight:
 draft: false
+disabled: true
 ---
--- a/hugo/content/en/docs/howto/constrain-min-items-list/index.md
+++ b/hugo/content/en/docs/howto/constrain-min-items-list/index.md
@@ -2,4 +2,5 @@
 title: How to constrain the minimum items in a list
 weight:
 draft: false
+disabled: true
 ---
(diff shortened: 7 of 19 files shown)


hugo/layouts/partials/paging/prev-next.html
```diff
@@ -4,8 +4,8 @@
 
 <div class="prevnext">
     <nav class="prevnext__container" aria-label="{{ T "ui_pager_name_prevnext" }}">
-        {{- if $prev -}}
-            <a class="prevnext__prev" href="{{ $prev.RelPermalink }}" aria-label="{{ T "ui_pager_prev" }} - {{ $prev.LinkTitle }}">
+        {{- if and $prev (ne $prev.Params.coming_soon true) -}}
```
JorindeUsMedia, line 7: You don't want to be stuck within the prev/next navigation here. Disabled pages should not be part of the prev/next . 
It should skip the prev/next pages that are disabled and go to the next page that is not disabled.

hugo/content/en/docs/howto/constrain-list-ints-no-duplicates/index.md
```diff
@@ -1,5 +1,6 @@
 ---
 title: How to constrain a list of integers to have no duplicates
-weight: 
+weight:
 draft: false
----
\ No newline at end of file
+coming_soon: true
```
JorindeUsMedia, line 5: I'm in doubt if we should call this coming_soon because it could be pages are disabled for another reason (doc outdated or something). Maybe just name this 'disabled' ?

JorindeUsMedia (changes requested):

hugo/content/en/docs/howto/constrain-list-ints-no-duplicates/index.md
JorindeUsMedia: @Lyoness What do you think is best content-wise? 'disabled' or 'coming_soon'. What's more clear for you?

Lyoness: @JorindeUsMedia yes, `disabled` is good.

hugo/layouts/docs/single.html
```diff
@@ -10,7 +10,11 @@
                         <h1 class="article__title">{{ .Title }}</h1>
                     </header>
                     <div class="article__content">
-                        {{ .Content }}
+                        {{ if .Params.coming_soon }}
+                            <div class="todo">this page will be coming soon</div>
```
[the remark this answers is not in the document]
Lyoness: > Not sure if we should force content when coming_soon is set to true.

So would this work like `draft` then? If set to true, the content doesn't show? 

I would treat the `disabled` param to mostly work for indexes.  A site visitor would see what types of content is intended to be written but hasn't yet.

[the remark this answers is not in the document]
JorindeUsMedia: @Lyoness No it doesn't work together with Draft=true. Hugo will skip all content with draft set to true so it won't show up anywhere. If you want a page to be disabled in the nav you have to set draft to false and disabled to true.

hugo/layouts/partials/paging/prev-next.html
JorindeUsMedia: For nowI think: don't disable next/prev. Disabled will show greyed out in docs. 
Would be nice if we can also skip disabled pages in the prev/next nav but I think that will be a lot of work because hugo doesn't have anything build in for this.

Lyoness: > I think it's best to not disable the prev/next: than you just land on a todo/coming-soon page. I think that's better than being stuck 

I like this solution.

hugo/layouts/docs/single.html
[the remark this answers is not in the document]
Lyoness: @JorindeUsMedia I think the above solution of having a link disabled in the nav, but when clicked on, it says, "TODO" is a good option.  We can tweak later if that doesn't work.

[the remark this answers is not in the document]
JorindeUsMedia: @Lyoness So just to summarize: Do you want us to not the markdown show content but show a todo when disabled is to true.
Or do want to have total control over this yourself? Which means we won't do anything in the detailpage itself  when disabled is set to true.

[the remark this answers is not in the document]
Lyoness: >not [show] the markdown show content but show a todo when disabled is to true. 

Let's do this for option and see how it works.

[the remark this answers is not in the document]
JorindeUsMedia: @anne-usmedia So for you: Keep what you have in single.html (but use 'disabled' instead of 'coming_soon') but make sure it comes from the translation files.

hugo/layouts/partials/paging/prev-next.html
```diff
@@ -5,7 +5,7 @@
 <div class="prevnext">
     <nav class="prevnext__container" aria-label="{{ T "ui_pager_name_prevnext" }}">
         {{- if $prev -}}
-            <a class="prevnext__prev" href="{{ $prev.RelPermalink }}" aria-label="{{ T "ui_pager_prev" }} - {{ $prev.LinkTitle }}">
+            <a class="prevnext__prev" href="{{ $prev.RelPermalink }}" aria-label="{{ T "ui_pager_prev" }} - {{ $prev.LinkTitle }}"}>
```
JorindeUsMedia, line 8: Issue with an extra } at the end

JorindeUsMedia (changes requested):

hugo/layouts/partials/paging/prev-next.html
anne-usmedia: fixed

Lyoness (approved):
Importing to Gerrit

Lyoness:
https://review.gerrithub.io/c/cue-lang/cuelang.org/+/550894

Lyoness:
Closes [f748861](https://github.com/cue-lang/cuelang.org/commit/f748861bfdc6af57f9f5f1b72b0e8ad8019c925c) as merged.

PR #319, opened 2023-03-08
Hugo: Add quote shortcode

anne-usmedia:
- Styling regular markdown quote
- Added quote shortcode
- Styling quote shortcode
- Added 300 weight font
- Added leading zeros in base style
- Changed lightest grey
- Added quote shortcode example page

For https://linear.app/usmedia/issue/CUE-122

PR #312, opened 2023-03-03
Add upcoming shortcodes to example page (/examples/)

anne-usmedia:
- Added disabled links for future shortcodes 
    - Carousel 
    - CTA 
    - Quote
- Created empty pages for the upcoming shortcodes 
    - Carousel 
    - CTA 
    - Quote 
    - Tabs
- Fixed the order/weights of the shortcodes with the upcoming shortcodes in mind

For [https://linear.app/usmedia/issue/CUE-159](https://linear.app/usmedia/issue/CUE-159)

PR #324, opened 2023-03-10
Hugo: Add CTA shortcode

anne-usmedia:
- Create CTA shortcode
- Adjust CTA styling
- Add dark-blue button
- Add CTA example page

For https://linear.app/usmedia/issue/CUE-120


hugo/assets/scss/components/cta.scss
```diff
@@ -2,49 +2,74 @@
...
     }
 
-    &--yellow {
-        background-color: $c-yellow;
+    &--flat {
+        --cta-color: #{ $c-blue };
+
+        align-items: normal;
+        border-radius: 5px;
+        gap: 0;
```
JorindeUsMedia, line 39: css order is not alphabetic

hugo/assets/scss/components/cta.scss
```diff
@@ -2,49 +2,74 @@
...
+        font-size: 1rem;
+        line-height: inherit;
```
JorindeUsMedia, line 41: is it needed to set to inherit? I thought that was default behaviour so maybe this line can be removed?

JorindeUsMedia (changes requested):

hugo/assets/scss/components/cta.scss
anne-usmedia: The lineheight of the flat differs from the regular CTA. The one from the regular CTA is smaller than the body text. The lineheight of --flat needs to be the same as the body. With inherit it gets the same lineheight of the body.

I can also give it the same value as the body lineheight, but than you have to change it in multiple places if it needs to be adjusted.

JorindeUsMedia: ah ok, I though inherit took the value of whatever parent with a line-height comes first. Misunderstood that :)

anne-usmedia:
@Lyoness same as with https://github.com/cue-lang/cuelang.org/pull/319, I will activate the link in this issue after https://github.com/cue-lang/cuelang.org/pull/312 is merged

Lyoness (approved):
Importing to Gerrit

Lyoness:
https://review.gerrithub.io/c/cue-lang/cuelang.org/+/550808

anne-usmedia:
@Lyoness just rebased this branch

myitcv:
Thanks - reimporting now.

PR #325, opened 2023-03-10
Hugo: Add missing css variables

JorindeUsMedia:
- Use css variables everywhere so they can be easily overwritten when we want a dark theme
- Use same way of using variables everywhere
- Add missing imports & mixin
- Add defaults for variables in case variables are not set or coming from outside current component
- Change hover state colors of blue button because difference between normal and hover state was too little

For: https://linear.app/usmedia/issue/CUE-130

### DIFF
--- a/hugo/assets/scss/base/base.scss
+++ b/hugo/assets/scss/base/base.scss
@@ -13,25 +13,31 @@
     --text-color: #{ $c-grey--darkest };
     --heading-color: #{ $c-blue };
     --text-link-color: #{ $c-blue };
+
     --caption-color: #{ transparentize($c-grey--darkest, 0.3) };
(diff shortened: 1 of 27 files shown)


Lyoness:
@JorindeUsMedia please can you rebase this one more time?

myitcv:
Thanks @JorindeUsMedia - I definitely don't fully grok all the CSS stuff, but the commit message made it easier to understand what you're up to here!

Importing into Gerrit now.

JorindeUsMedia:
@myitcv Rebased and fixed the conflicts

myitcv:
Closed via ccba4785e030ea701f57658a811bc6723f6f843d

PR #328, opened 2023-03-15
Hugo/Editor: Styling alignment with design

anne-usmedia:
Adjusted and/or changed the styling site-wide
to align with the newest designs.

For https://linear.app/usmedia/issue/CUE-171


editor/src/scss/base/fonts.scss
```diff
@@ -1,16 +1,25 @@
...
+// Inter (latin1, 300, 400, 500, 600, 700)
 // https://gwfh.mranftl.com/fonts/inter
 //
 
+@font-face {
+    font-display: swap;
+    font-family: 'Inter';
+    font-style: normal;
+    font-weight: 300;
+    src: url('/fonts/inter/300.woff2') format('woff2'), /* Chrome 36+, Opera 23+, Firefox 39+ */
```
JorindeUsMedia, line 11: Path should be /static/fonts/ here. Same for all other paths within this file

hugo/assets/scss/components/separator.scss
```diff
@@ -10,8 +11,9 @@
     clear: both;
     display: block;
     height: 1px;
-    margin: 1rem auto;
+    margin: 0.5rem auto 1rem;
     max-width: $w-content - ($p-gutter * 2);
+    opacity: 0.2;
```
JorindeUsMedia, line 16: Double check if this is working correctly because I remember the hr tag itself also has opacity because it has a rgba background set. We want this set or by opacity or by rgba, not both I think

hugo/assets/scss/components/sidenote.scss
```diff
@@ -15,7 +15,7 @@
     font-weight: $weight-medium;
     height: var(--sidenote-height);
     line-height: var(--sidenote-height);
-    margin: 0 0 .5em;
+    margin: 0 0 0.5em;
```
JorindeUsMedia, line 18: is there a reason we use em here instead of rem? rem depend on the base font size (16px in this case). Em use the font-size of the current element as base

hugo/assets/scss/mixins/scrollbar.scss
```diff
@@ -10,18 +10,20 @@
...
 
         &:hover {
             background-color: var(--scrollbar-color-hover, #{ $c-scrollbar-color-hover });
         }
     }
 
     // Standard version (Firefox only for now)
-    scrollbar-color: var(--scrollbar-color, #{ $c-scrollbar-color }) var(--scrollbar-background, #{ $c-scrollbar-background });
+    scrollbar-color:
+        var(--scrollbar-color, #{ $c-scrollbar-color })
```
JorindeUsMedia, line 27: default value can be without interpolation ( #{} )   I think

JorindeUsMedia (changes requested):

hugo/assets/scss/components/sidenote.scss
JorindeUsMedia: Just checked since I did this part of the styling I think: we can this to rem because we use rem everywhere. It won't affect the styling here.

hugo/assets/scss/components/separator.scss
anne-usmedia: Just checked this. We use transparentize to set the `--hr-color`. I don't know if it works on the `hr` but it doesn't changes anything on the separator. Opacity does lighten the separator. How do you want this fixed?

JorindeUsMedia: I thought separator was set on an hr, but it seems it's a div. In that case this is fine. Sorry for the confusion :sweat_smile:

JorindeUsMedia (approved):

myitcv:
Importing to Gerrit now.

myitcv:
Closed via 1c05568f44a694b16b47153ac797aae882baac5d

PR #329, opened 2023-03-15
Hugo: TS lint & TS config issues

JorindeUsMedia:
- Set noJSConfigInAssets to true in hugo config because we don't want a jsconfig file since we're using typescript. Jsconfig file sometimes 'wins' from tsconfig in some editors which causes validation issues.
- Remove jsconfig.json from hugo assets
- Update tsconfig to use newer supported es version so we can use es2020 in our ts files but also makes compilied version more efficient because we don't need to support really old browsers anymore (IE etc)
- Fix linter issues with any typing in main.ts for datalayer. For now ignore this line since dataLayer is external and not in use yet
- Fix linter issue & ts issue with queryselectorALl & loop

For: https://linear.app/usmedia/issue/CUE-180


anne-usmedia (approved):

myitcv (approved):
LGTM - but really deferring to @anne-usmedia's review when I say that :)

myitcv:
Importing into Gerrit now.

myitcv:
Closed via 1fce05543d9682d327771f25da56f4fff459059f

PR #330, opened 2023-03-15
Hugo: Fix scrolling issues in codeblock

JorindeUsMedia:
- Move highlight config from config.toml to markup.toml
- Make sure horizontal overflow doesn't get wrapped but gets a scrollbar
- Make sure vertical overflow for pre doesn't get cut of but shows all content. Remove height: 100% from css because this only works when pre or highlight is the only child which is not the case.
- Use flex to make last highlight or pre tag in a column fill out the remaining space.

For: https://linear.app/usmedia/issue/CUE-177 and
     https://linear.app/usmedia/issue/CUE-176
     
   
Test urls:
https://deploy-preview-330--cue.netlify.app/examples/basic/block#code-blocks
https://deploy-preview-330--cue.netlify.app/docs/tour/validation/
https://deploy-preview-330--cue.netlify.app/docs/tour/definitions/

### DIFF
--- a/hugo/assets/scss/base/base.scss
+++ b/hugo/assets/scss/base/base.scss
@@ -346,13 +346,10 @@ pre {
     border: 1px solid var(--pre-border-color);
     border-radius: 5px;
     color: var(--pre-color);
-    height: 100%;
     margin: 0 0 1rem;
     max-width: 100%;
-    overflow: auto;
+    overflow-x: auto;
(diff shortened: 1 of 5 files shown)


myitcv (approved):
LGTM in terms of the visual fixes!

Will import into Gerrit now.

myitcv:
Closed via b9486b40b9566e1b12b0edb07035c29a3706c67d

PR #333, opened 2023-03-16
Hugo: show correct hover for multiple line texts

anne-usmedia:
When the hovered text is more than 1 line, show the hover animation on all the lines, not only the last

For https://linear.app/usmedia/issue/CUE-182

### DIFF
--- a/hugo/assets/scss/components/nav.scss
+++ b/hugo/assets/scss/components/nav.scss
@@ -42,7 +42,6 @@
 
     &__link {
         cursor: pointer;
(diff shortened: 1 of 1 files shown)


JorindeUsMedia (approved):

myitcv (approved):
LGTM. Tested locally to verify that https://review.gerrithub.io/c/cue-lang/cuelang.org/+/551000 when applied on top of this change fixes the issue reported in https://linear.app/usmedia/issue/CUE-182

PR #338, opened 2023-04-11
Hugo: Shortcode Tabs

JorindeUsMedia:
- Add new shortcode for tabs & tab. Tab needs to be used inside tabs
- Add html and styling for tabs
- Add javascript to tab-group do determine when we show pagination controls for when tabs don't fit inside the current container or screen, handle click events on next/prev buttons which triggers the scrolling of the tabgroup & add disabled attribute when prev/next is not needed because you can't scroll any further.
- Add javascript to handle click events on tabs. Save tab selection when group is not the default group and sync across tabs of the same group. Also save selection to localstorage and restore the tabs selection on page load.
- Add example page for tabs.

For: https://linear.app/usmedia/issue/CUE-83

Preview: https://deploy-preview-338--cue.netlify.app/examples/shortcodes/tabs/

### DIFF
--- a/hugo/assets/scss/components/tabs-nav.scss
+++ b/hugo/assets/scss/components/tabs-nav.scss
@@ -1,0 +1,144 @@
+@import '../config/colors';
+@import '../config/sizes';
+@import '../mixins/stretch';
+@import '../mixins/typography';
+
+.tabs-nav {
+    $self: &;
+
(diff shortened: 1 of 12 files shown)


myitcv (approved):
The end result LGTM - I haven't got a clue about the TypeScript, CSS etc :)

Will import to GerritHub now.

PR #339, opened 2023-04-12
Feat: Add category overview chapters to docs

JorindeUsMedia:
- Add layout for category overviews which gets all pages within parent with same category as set in front-matter of the overview page and render this on the page below the content.
- Add categories to doc detail pages
- Add logic to navtree to render pages of category-chapter different because they are not nested within the section. Also make sure docs that have categories in the front-matter are not rendered in the nav tree as part of the
 normal doc flow because otherwise they will be in there twice. Note: this
 means we assume docs with a category set will also have an overview-page
 set up for that category

 Note: Hugo taxonomies (categories and tags) are still turned off. We don't
 need to turn them on for now because we only use categories in a custom way.
 Turning them on means Hugo will also generate overview pages for each
 category on a top level (/categories/language for instance) and I think we
 don't want that now.

For https://linear.app/usmedia/issue/CUE-179

Preview: https://deploy-preview-339--cue.netlify.app/docs/howto/language/
https://deploy-preview-339--cue.netlify.app/docs/howto/list-no-duplicates/

### DIFF
--- a/hugo/content/en/docs/howto/ensure-min-max-list/index.md
+++ b/hugo/content/en/docs/howto/ensure-min-max-list/index.md
@@ -2,6 +2,8 @@
 title: Require min or max items in a list
 weight:
 draft: false
+categories:
+    - Language
 ---
 
 ## Min
--- a/hugo/content/en/docs/howto/language/index.md
+++ b/hugo/content/en/docs/howto/language/index.md
@@ -1,0 +1,7 @@
+---
+title: Language
+weight:
+draft: false
+layout: category-overview
+category: Language
+---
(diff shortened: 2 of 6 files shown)


myitcv:
@anne-usmedia helping to rebase here.

Looks good @JorindeUsMedia - thanks for the commentary in the PR description/commit message.

One question: why do we still have docs appearing at the "top level" of Howtos in the LHS nav, as siblings of the "Language" category?

Screenshot 2023-04-13 at 13 40 25

myitcv (approved):
LGTM for now. @JorindeUsMedia we can address my question in a follow up as required.

myitcv:
Merged via 2d79ae6a17862d08f6e5f51b8813c7192e4fc1f0

PR #344, opened 2023-04-17
Hugo: Code component v2: Code-tabs

JorindeUsMedia:
- Move tabs code to partials so we can reuse it for both tabs & code-tabs
- Adjust rendering of code fences to have new style
- Use Roboto Mono as monospace font instead of browser default.
- Style inline code tag different from pre & highlight block.
- Code tabs component grid html & styling _ Different styling for Tabs & Tabs-nav component with code modifier
- Style inner code-blocks & pre tags different when inside tabs or code-tabs.
- Add example page for code-tabs
- Move basic code-block example to separate page because basic-blocks page got too long

For: https://linear.app/usmedia/issue/CUE-200 and
https://linear.app/usmedia/issue/CUE-199


myitcv (approved):
This looks _so_ much better, thank you!

Importing to GerritHub now. We can always tweak bits here and there later.

But landing this means I can work on integrating the preprocessor with this new approach.

myitcv:
Ah, just spotted that this needs a rebase, please @JorindeUsMedia!

JorindeUsMedia:
@myitcv Rebasing is done :)

PR #347, opened 2023-05-10
Hugo: Add Algolia indexing

anne-usmedia:
- Add config for algolia.json
- Add Atomic Algolia package + script. This sends the algolia.json to Algolia
- Add list.algolia.json to create the algolia.json
- Add .env sample; needed values can be found in Algolia
- Set frontmatter of community page to yaml (was the only one in toml)
- Update readme for Algolia testing

### DIFF
--- a/hugo/README.md
+++ b/hugo/README.md
@@ -67,6 +67,23 @@ If all above is done and the icon is added in the right folder you can run the f
 npm run icons
 ```
 
+### Algolia
+
+If you want to test Algolia locally, follow the next steps:
+
+1. Have (admin) access to an Algolia account
+2. Create a test index in Algolia
+3. Find your Algolia API keys (settings > Team and Access > API Keys),
+   fill the keys in the following command,
+   and run the command
+
+```
+ALGOLIA_APP_ID={{ YOUR_APP_ID }} ALGOLIA_ADMIN_KEY={{ YOUR_ADMIN_KEY }} ALGOLIA_INDEX_NAME={{ YOUR_INDEX_NAME }} ALGOLIA_INDEX_FILE="../_public/algolia.json" npm run algolia
+```
+
+4. Do your testing
+5. Delete the test index
+
 ## History
 
 This site was setup using the following guides:
--- a/hugo/layouts/_default/list.algolia.json
+++ b/hugo/layouts/_default/list.algolia.json
@@ -1,0 +1,55 @@
+{{/* Generates a valid Algolia search index */}}
+{{- $section := $.Site.GetPage "section" .Section -}}
+[
+    {{ range $idx, $page := $.Site.Pages -}}
+        {{- if or (and ($page.IsDescendant $section) (and (not $page.Draft) (not $page.Params.no_index) (not (eq $page.Type "examples")))) $page.IsHome -}}
+            {{- if $idx -}},{{- end -}}
+
+            {{- template "algolia-object" (dict
+                "index" 0
+                "page" $page
+                "section" $section
+            ) -}}
+        {{- end -}}
+    {{- end }}
+]
+
+{{- define "algolia-object" -}}
+    {{ $page := .page -}}
+    {{ $cutoff := 1000 -}}
+    {{ $index := .index -}}
+    {{ $skipWords := mul $cutoff $index }}
+
+    {{- $objectID := (print $page.File.UniqueID "_" (add $index 1)) -}}
+    {{- $title :=  $page.Title -}}
+    {{- $link := $page.Permalink -}}
+    {{- $categories := cond (isset $page.Params "categories") ($page.Params.categories) "" -}}
+    {{- $order := (add $index 1) -}}
+    {{- $publishDate := $page.PublishDate -}}
+    {{- $summary := $page.Summary -}}
+    {{- $content := "" -}}
+
+    {{- if eq $index 0 -}}
+        {{- $content = delimit (first $cutoff $page.PlainWords) " " -}}
+    {{- else -}}
+        {{- $content = delimit (first $cutoff (after $skipWords $page.PlainWords)) " " -}}
+    {{- end -}}
+
+    {
+        "objectID": {{ $objectID | jsonify }},
+        "order": {{ $index | jsonify }},
+        "title": {{ $title | jsonify }},
+        "categories": {{ $categories | jsonify }},
+        "link": {{ $link | jsonify }},
+        "publishDate": {{ $publishDate | jsonify }},
+        "summary": {{ $summary | jsonify }},
+        "content": {{ $content  | jsonify}}
+    }
+
+    {{- if gt $page.PlainWords (add $cutoff $skipWords) -}},
+        {{- template "algolia-object" (dict
+            "index" (add $index 1)
+            "page" $page
+        ) -}}
+    {{- end -}}
+{{- end -}}
--- a/hugo/package.json
+++ b/hugo/package.json
@@ -7,6 +7,7 @@
   "license": "ISC",
   "dependencies": {
     "algoliasearch": "4.17.0",
+    "atomic-algolia": "0.3.19",
     "mathjax": "3.2.2",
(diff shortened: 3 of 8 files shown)


anne-usmedia:
There are 2 more things that need to be done:
1. Add the Algolia variables to Github
2. Add the `algolia` script to the pipeline

hugo/README.md
```diff
@@ -67,6 +67,16 @@ If all above is done and the icon is added in the right folder you can run the f
...
 
+### Algolia
+
+If you want to test Algolia local, you can add the environoment variables from Algolia to the environment config:
+
+```
+cp .env.sample .env
+```
+
+The values can be found in Algolia. For testing, create a new index. DO NOT USE THE cuelang.org INDEX.
```
JorindeUsMedia, line 78: Maybe we can allready create a test index for this that can be used. You have a maximum of index so we don't want everybody to just create new indexes

myitcv: How is Algolia tested locally?

> You have a maximum of index so we don't want everybody to just create new indexes

What does this mean?

hugo/layouts/_default/list.algolia.json
```diff
@@ -0,0 +1,53 @@
...
+        "order": {{ $index | jsonify }},
+        "title": {{ $title | jsonify }},
+        "categories": {{ $categories | jsonify }},
+        "link": {{ $link | jsonify }},
+        "publishDate": {{ $publishDate | jsonify }},
+        "summary": {{ $summary | jsonify }},
+        "content": {{ $content  | jsonify}}
+    }
+    
+    {{- if gt $page.PlainWords (add $cutoff $skipWords) -}},
```
myitcv, line 47: I thought you mentioned that one of our pages trips the size limit on the index?

Where is the distinct setting applied?

https://www.algolia.com/doc/guides/sending-and-managing-data/prepare-your-data/how-to/indexing-long-documents/

hugo/package.json
```diff
@@ -6,6 +6,7 @@
   "author": "",
   "license": "ISC",
   "dependencies": {
+    "atomic-algolia": "0.3.19",
```
myitcv, line 9: This appear to be a somewhat random package. Is there a reason not to use an official Algolia client?

hugo/layouts/_default/list.algolia.json
```diff
@@ -0,0 +1,53 @@
+{{/* Generates a valid Algolia search index */}}
+{{- $section := $.Site.GetPage "section" .Section }}
+[
+    {{ range $idx, $page := $.Site.Pages -}}
+    {{- if $idx -}},{{- end -}}
+        {{- if or (and ($page.IsDescendant $section) (and (not $page.Draft) (not $page.Params.private))) $section.IsHome -}}
```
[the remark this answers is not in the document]
myitcv: Indeed. All of the `/examples` pages feature in the current index (generated locally), as does the search page. Those should not be indexed.

hugo/layouts/_default/list.algolia.json
```diff
@@ -0,0 +1,53 @@
...
+    {{- $title :=  $page.Title -}}
+    {{- $link := $page.Permalink -}}
+    {{- $categories := cond (isset $page.Params "categories") ($page.Params.categories) "" -}}
+    {{- $order := (add $index 1) -}}
+    {{- $publishDate := $page.PublishDate -}}
+    {{- $summary := $page.Summary -}}
+    {{- $content := "" -}}
+
+    {{- if eq $index 0 -}}
+        {{- $content = delimit (first $cutoff $page.PlainWords) " " -}}
```
myitcv, line 31: The current approach simply appears to simply strip markdown/hugo markup from the input. However this includes code blocks and their output in the index. Is this intentional?

hugo/README.md
```diff
@@ -67,6 +67,16 @@ If all above is done and the icon is added in the right folder you can run the f
 npm run icons
 ```
 
+### Algolia
+
+If you want to test Algolia local, you can add the environoment variables from Algolia to the environment config:
+
+```
+cp .env.sample .env
```
myitcv, line 75: Not everyone uses `.env` files - I would simply document what needs to be done to run Algolia locally

myitcv (commented):
Happy to help get this incorporated when it's ready.

myitcv (commented):
Happy to help get this incorporated when it's ready.

hugo/package.json
JorindeUsMedia: This package uses the algolia client. As far as I know algolia client does have enpoints you can call but you still have to do the logic whether to update, add, delete somethings yourself. This package loops through the json and checks if an entry needs to be added, updated or deleted from algolia. It's a small package so we chose to use this instead of building the same thing ourselves

hugo/layouts/_default/list.algolia.json
JorindeUsMedia: Yes we're currently using hugo's plain function. This strips all html and markdown and keeps the raw content including everything inside shortcodes. People might be searching on code pieces? They will also show up. If we don't want this we have to look into building a custom hugo function but I'm not sure that's possible in hugo. (https://discourse.gohugo.io/t/adding-custom-functions/14164)

hugo/layouts/_default/list.algolia.json
JorindeUsMedia: We have set something up for that as algolia told us but inside algolia it didn't look like expected. When we start using algolia search enpoint we have to see if we set it up correctly and check what results it gives us back. I'm not sure if Anne for now turned if off to be able to see if indexing worked correctly with the partial page entries

hugo/README.md
JorindeUsMedia: Maybe we should just refrase this sentence. If you want to test something with algolia you don't want to overwrite the current index because that would mess up search results on alpha.So if we need to add something for instance and we need to test is we have to create a test index to play around with it without changing the production index. So we could say that a develop who needs to test something with algolia has to create a new index just for testing and after they're done they should remove it or something so we don't end up with 'dead' test indexes

hugo/layouts/_default/list.algolia.json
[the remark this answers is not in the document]
JorindeUsMedia: @anne-usmedia We could skip everything in examples in this loop itself probably.
And we can add an attribute in the front-matter: no_index or noIndex (not sure what case we use now) and if that is set skip the entry

hugo/package.json
myitcv: Why not use the Algolia CLI? 

https://github.com/algolia/setup-algolia-cli

JorindeUsMedia: @anne-usmedia Maybe you can have a look if it's easy to use that one instead of atomic-algolia?

hugo/README.md
anne-usmedia: Reworded this into a roadmap

hugo/layouts/_default/list.algolia.json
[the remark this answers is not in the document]
anne-usmedia: All drafts, example pages, and no_index pages are skipped. I added a `no_index: true` to the search results page. If there are more pages that shouldn't be included in the search results, the `no_index: true` can be added in the frontmatter

hugo/package.json
anne-usmedia: Discussed with @myitcv. Since the CLI makes it hard to delete records, we are going to use Atomic for now. If needed, we can build it ourselves in the future, but for now this will do.

myitcv (approved):
LGTM. Importing into Gerrit.

myitcv:
Closed via d724b05e692e059e5a3f55ad9db4d815c526eec8

myitcv:
> I will do this in a follow-up change.

Just awaiting a review of https://review.gerrithub.io/c/cue-lang/cuelang.org/+/554299 from @anne-usmedia or @JorindeUsMedia

PR #348, opened 2023-05-12
Hugo: Fill and style search result page with algolia

anne-usmedia:
- Add Algolia Search package
- Restyle search bar and page
- Remove Google Search
- Add Algolia logo (mandatory with free plan)
- Add search widget for getting the results from Algolia
- Add search teaser styling

For feature/cue-210-search-redo-search-page-replace-google


hugo/assets/scss/components/teaser.scss
```diff
@@ -58,13 +59,46 @@
...
+            font-style: italic;
+            font-weight: $weight-medium;
+            margin-top: 0.5rem;
+        }
+    }
+
+    @include screen($screen-normal) {
+        &--search {
+            #{ $self }__container {
+                padding: 2rem 20px 20px;
```
JorindeUsMedia, line 97: maybe use the p-gutter variable intead of 20px?

anne-usmedia: done

hugo/assets/ts/widgets/search.ts
```diff
@@ -0,0 +1,117 @@
...
+            link: hit.link,
+            categories: hit.categories,
+            summary: hit._snippetResult.summary.value,
+        };
+    }
+
+    public createTeaser(teaser: Teaser): string {
+        return `
+            <div class="teaser teaser--search">
+                <div class="teaser__container">
```
JorindeUsMedia, line 87: why do we need a container wrapper? is it possible to add padding styles on the teaser itself?

anne-usmedia: Because that is how the OG teaser was set up. So of course that is possible, but than we have to change to original teaser.html as well.

JorindeUsMedia:
Would be nice to have even padding at top and bottom. Now there's more padding at the top of the teaser. When there is a summary you won't notice this because than the p tag also has a margin. Maybe it's better to remove the margin on the summary and make the padding on the teaser top and bottom even. Also maybe we can get rid of the __container bem and move these styles to teaser itself? Or is there a reason we need 2 div's for this?

Screenshot from 2023-05-22 15-24-07

hugo/assets/ts/interfaces/search.ts
```diff
@@ -0,0 +1,9 @@
+export interface SearchItem {
+    objectID: string;
+    title: string;
+    link: string;
+    categories: string[];
```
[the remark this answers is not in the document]
anne-usmedia: Same goes for category (and when that one is emty, for summary as well). When there is no data, the attribute is created anyway. So it is always there, but might be an empty string.

hugo/assets/ts/widgets/search.ts
anne-usmedia: Container class is removed in both the teaser.html and search teaser & styling is moved to teaser class

hugo/assets/ts/interfaces/search.ts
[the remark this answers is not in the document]
JorindeUsMedia: if categories is empty will it return an empty string or an empty array?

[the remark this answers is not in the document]
anne-usmedia: Hmm a string. So `string | string[]`?

JorindeUsMedia (approved):

myitcv:
@JorindeUsMedia @anne-usmedia - is this waiting on https://github.com/cue-lang/cuelang.org/pull/347?

anne-usmedia:
@myitcv not per se, the search results you get are the results that are in Algolia. For now these are the results I got working on #347 (locally) though, and not the results from the live environment.

myitcv:
> @myitcv not per se, the search results you get are the results that are in Algolia. For now these are the results I got working on #347 (locally) though, and not the results from the live environment.

@anne-usmedia - just to confirm, is this the result that clicking on a search result in this PR links to https://localhost:1313/docs/language-guide/data/bytes-values/?

myitcv:
Closed via f3fdba77476453e783ab48ce5f9acd0b9a10bbc8. 

We should look to get #347 merged asap because the search results are currently broken linking to localhost.

anne-usmedia:
@myitcv yes, since those are the links known by Algolia

PR #351, opened 2023-05-23
Playground - Restructure main react app

JorindeUsMedia:
- Split application in multiple folders. Containers for page components (in future the policy mode will also become a seperate container), components for regular components, config for config like cue version, models for types & interfaces
- Add tsconfig paths plugin for webpack & define path aliases in tsconfig so we can make imports shorter
- Change way of handling inputs from array to object, so we can use named keys instead of indexed based which was a bit difficult to understand what was going on.
- Change way of handling dropdown inputs. Remove input classes. Move handling of updating url to the app in 1 handler function instead of per input to improve readability and make 1 function responsible for updating the url instead of multiple.
- Add types for inputs. Move these to a separate file instead of putting everything in app.tsx
- Fix issues with state handling because setState can take some time to update: Use the callback function on setState where needed.
- Dropdown: Allow both links & buttons with a callback function
- Dropdown: Split buttonTitle into prefix & disabled text props.
- Dropdown: Keep track of activeValue inside the dropdown itself as part of the state so it immediately updates without delay. State could take a while to update so it could take some time after clicking an option before the
 dropdown button itself updated. This is fixed by using activeValue as part of
  dropdown state.
 - Rename some values and functions so they better describe what is done.

For: https://linear.app/usmedia/issue/CUE-218


myitcv (approved):
LGTM will import into Gerrit now.

myitcv:
Closed via 44d51ed27549e0e22b900b3ceb578480d2ce991f

PR #358, opened 2023-06-01
Hugo: add copy button to code tabs

anne-usmedia:
- Add copy button to code tabs
- Add copy functionality
- Add styling for copy button + success message

For https://linear.app/usmedia/issue/CUE-166


hugo/assets/scss/components/tabs.scss
```diff
@@ -28,6 +29,43 @@
...
+        display: block;
+        padding: 4px 8px;
+        position: absolute;
+        right: 0;
+        top: 35px;
+        width: fit-content;
+        z-index: 1;
+
+        &::before {
+            border-color: transparent transparent #fdc52f transparent;
```
JorindeUsMedia, line 44: Move hex value to a scss variable

hugo/assets/ts/widgets/copy.ts
```diff
@@ -0,0 +1,61 @@
...
+        elements.forEach((element) => {
+            const newWidget = new Copy(element);
+            newWidget.init();
+        });
+    }
+
+    public init(): void {
+        this.copyButtons.forEach(button => {
+            button.addEventListener('click', () => {
+                const content: string = this.element.querySelector<HTMLElement>('.tabs__item.is-active').textContent;
```
JorindeUsMedia, line 33: you now get the content en element on every click. Maybe move them outside the click event (so 2 lines up, within the foreach, before the click)

hugo/assets/ts/widgets/copy.ts
```diff
@@ -0,0 +1,61 @@
...
+                const messageElement: HTMLElement = button.parentElement.querySelector('.tabs__message');
```
JorindeUsMedia, line 34: same here, see comment above

hugo/layouts/shortcodes/code-tabs.html
```diff
@@ -6,7 +6,7 @@
 {{- $tabsBottomLeft := .Scratch.Get "tabs-bottom-left" -}}
 {{- $tabsBottomRight := .Scratch.Get "tabs-bottom-right" -}}
 
-<div class="code-tabs">
+<div class="code-tabs" data-copy>
```
JorindeUsMedia, line 9: you want to copy the content of the current tab so I think data-copy should be on a nested element, probably the code block itself?

JorindeUsMedia (changes requested):
Styling comments:
Icon is a bit dark now and draws more attention than the code itself: use a lighter grey.
Success message font is a bit big compared to code and tabs: use smaller font?

JorindeUsMedia (approved):

myitcv (changes requested):
Thanks. A couple of things I think we need to change here:

1. When we previously discussed this, we talked about the copy button not always being visible, but instead only appearing when the mouse is over the thing that can be copied. For example, see https://github.com/cue-lang/cue/blob/master/doc/contribute.md#asserting-a-developer-certificate-of-origin in an incognito window (for some reason the new-style layout always shows the copy button, which I think must be a bug)
2. Perhaps it's just me (let me know what you think) but it feels like the copy button should be within the tab, not in the tab header.
3. Please can we emulate the GitHub behaviour when the button is pressed? Changing the icon to a tick (with green border and colour), making the "Copied!" text appear to the left (makes most sense given the mouse might obstruct the text below), and using a more subtle dark grey instead of yellow. 
4. In most situations the contents of the tab will be exactly what should be copied, but not all. We need to handle situations where there are commands in a code block, and only copy the commands and not the `$ ` prefix. Only the preprocessor can know how to do this, so we need to somehow pass this text to be copied to the block in some way.

Regarding point 4, the way I have done this in the past is to render one thing, but pass a base64-encoded version of the text to be copied to a component via an attribute. That way, when the button is pressed, the attribute value is base64 decoded, and used as a the copy text, instead of what is rendered. I'm open to ideas, but something like that could work here. 

Happy to jump onto a call later to discuss.

JorindeUsMedia:
> Thanks. A couple of things I think we need to change here:
> 
> 1. When we previously discussed this, we talked about the copy button not always being visible, but instead only appearing when the mouse is over the thing that can be copied. For example, see https://github.com/cue-lang/cue/blob/master/doc/contribute.md#asserting-a-developer-certificate-of-origin in an incognito window (for some reason the new-style layout always shows the copy button, which I think must be a bug)
> 2. Perhaps it's just me (let me know what you think) but it feels like the copy button should be within the tab, not in the tab header.
> 3. Please can we emulate the GitHub behaviour when the button is pressed? Changing the icon to a tick (with green border and colour), making the "Copied!" text appear to the left (makes most sense given the mouse might obstruct the text below), and using a more subtle dark grey instead of yellow.
> 4. In most situations the contents of the tab will be exactly what should be copied, but not all. We need to handle situations where there are commands in a code block, and only copy the commands and not the `$ ` prefix. Only the preprocessor can know how to do this, so we need to somehow pass this text to be copied to the block in some way.
> 
> Regarding point 4, the way I have done this in the past is to render one thing, but pass a base64-encoded version of the text to be copied to a component via an attribute. That way, when the button is pressed, the attribute value is base64 decoded, and used as a the copy text, instead of what is rendered. I'm open to ideas, but something like that could work here.
> 
> Happy to jump onto a call later to discuss.

All of the feedback above has been implemented

myitcv:
Closed via 76cfc08539fa38f11b8040759d29ed7de9fab154

myitcv (approved):
LGTM. Applied rebase after import.

myitcv:
Forgot to hit "submit" on the review before merging in GerritHub!

PR #361, opened 2023-06-07
Playground: Monaco Editor

JorindeUsMedia:
- Remove ace-editor and replace with monaco editor
- Put config files for input/output  options editor in config folder
- On window resize recalculate editor sizes using layout() function
- Remove loadEditor function, move settings option to the editor config files.
- Move getting id and loading code from database to its own function
- Add css loader for .css files so Monaco styles can be loaded
- Overrule some css variables for monaco-editor to be able to change the background color of the output editor. Monaco only support 1 theme for the entire page/app so we can't use Monaco Theme's to change 1 editor only.
- Use vscode-light theme white for tabs backgrounds so the white is the same as the vscode-theme.
- Wait until editor loaded to get saved Code + update output
- Add deboucing to the inputhandler on the input-editor to avoid editor getting super slow because it has to handle each typed character.
- Fix issue with tabs content items missing a key which react wants to have.

### DIFF
--- a/playground/package-lock.json
+++ b/playground/package-lock.json
@@ -8,14 +8,16 @@
       "name": "CUEplayground",
       "version": "0.1.0",
       "dependencies": {
+        "@monaco-editor/react": "4.5.1",
+        "lodash.debounce": "4.0.8",
         "react": "18.2.0",
         "react-dom": "18.2.0",
         "react-scripts": "5.0.1"
       },
       "devDependencies": {
+        "@types/lodash.debounce": "4.0.7",
         "@types/react": "18.0.28",
         "@types/react-dom": "18.0.10",
-        "ace-builds": "1.15.0",
         "classnames": "2.3.2",
         "copy-webpack-plugin": "11.0.0",
         "css-loader": "6.7.3",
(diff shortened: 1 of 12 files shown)


myitcv:
> Add deboucing to the inputhandler on the input-editor to avoid editor getting super slow because it has to handle each typed character.

Why does this end up being necessary? From the WASM perspective, isn't the "cost" the same as before? Why does using Monaco change things here?

Only one request I would make here is to use a similar sized font to what we have currently. On my large screen, things look quite small in the editor panes. Unless this was done for another reason?

JorindeUsMedia:
@myitcv I also had some issues with it being slow using the old editor. I'm mainly working in firefox so could be that it's a firefox issue.. But I think some debouncing can't hurt?

About font size: Monaco editor fontsize is 16px. For the ace editor we used 15px so it's even a bit bigger than the previous editor

myitcv:
> I also had some issues with it being slow using the old editor. I'm mainly working in firefox so could be that it's a firefox issue.. But I think some debouncing can't hurt?

I would therefore prefer that we do that separately from changing the editor. The two changes should be entirely orthogonal (hence my question). 

> About font size: Monaco editor fontsize is 16px. For the ace editor we used 15px so it's even a bit bigger than the previous editor

This is what I see on my local machine. On the LHS, the current alpha.cuelang.org. On the RHS, the proposed monaco editor version.

Screenshot 2023-06-09 at 13 55 49

The RHS version is noticeably smaller to my mind, unless I'm missing something?

JorindeUsMedia:
@myitcv Hm that's weird. In my browser it looks like this
Left current alpha, right this PR preview (monaco)
Screenshot from 2023-06-09 15-35-01

In your browser the old one looks way bigger..

JorindeUsMedia:
@myitcv It seems mac & linux chrome render it differently. On Mac the old editor font looks way bigger because it shows a different font than on Linux. On Mac old editor uses Monaco Font 15px and on Linux it uses Ubuntu Mono 15px which is smaller than the Monaco font on mac (not to be confused with Monaco editor..)

JorindeUsMedia:
Just checked a windows machine: roughly same size as on Linux. So it seems just Mac renders the old editor font larger.

myitcv:
> Just checked a windows machine: roughly same size as on Linux. So it seems just Mac renders the old editor font larger.

So do we have a handle on what's going on here? 

i.e. everything is "ok" in the new setup?

myitcv:
The test failure is because of the issue I am fixing elsewhere.

myitcv:
The build will be fixed when this change lands: https://review.gerrithub.io/c/cue-lang/cuelang.org/+/555174

myitcv (approved):
LGTM - will import into GerritHub. We can deal with the playground icons issue later.

myitcv:
Closed via d8be17ccfda5b3b97d4e611e7bbf3ae1b2ab1211

PR #362, opened 2023-06-07
Hugo/Playground: Fix playground icons

JorindeUsMedia:
- Use absUrl pipe on hugo side to get the absolute url of the site which is needed to render the icons from the svg sprite.

Note: Local and preview did not have this problem, just production (alpha). 
Hopefully this fixes it and it will now get the absolute baseUrl. If not we have to look into the hugo variables for prod.


myitcv (approved):
LGTM. Will import into Gerrit.

myitcv:
Closed via 4af1457c1510d453c0f9d9ba7e50aaa53813dbf0

PR #364, opened 2023-06-20
Playground: Fix playground icons

JorindeUsMedia:
- Absolute url can have a trailing slash depending on the environment (local, preview, alpha) so added an extra check to see if we need a slash between baseUrl and img path in the playground spritesheet url.


myitcv (approved):
LGTM. Will import into GerritHub now.

PR #365, opened 2023-06-20
Playground: Share is not working

JorindeUsMedia:
- Fix issue with url containing a quote character
- Change urlPrefix function to return the whole share url so we have the url set in 1 place.
- Add check in response of getting shared code so the playground doesn't try to render the response when it's not successful.
- Change let contents to const because value never gets reassigned
- Remove unnecessary const for req because it's never used.

For: https://linear.app/usmedia/issue/CUE-241


myitcv (approved):
LGTM. Will import into GerritHub now.

myitcv:
Closed via 70da91066393a882f66af7db391f85e9fd3a05d8

PR #366, opened 2023-06-21
Search - Add more search attributes to index object

JorindeUsMedia:
- Add search attribute for tags & section (blog/docs)
- Change order of object keys so 'filters' are last

For: https://linear.app/usmedia/issue/CUE-235

### DIFF
--- a/hugo/assets/ts/interfaces/search.ts
+++ b/hugo/assets/ts/interfaces/search.ts
@@ -2,8 +2,10 @@ export interface SearchItem {
     objectID: string;
     title: string;
     link: string;
-    categories: string | string[];
     summary: string;
     publishDate: string;
     content: string;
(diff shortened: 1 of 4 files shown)


myitcv:
Just checking per my offline question: can Algolia search filter based on these three attributes?

JorindeUsMedia:
> Just checking per my offline question: can Algolia search filter based on these three attributes?

After this gets merged and the index gets updated I have to set them as searchable attributes & attributes for faceting in the algolia settings. Can do this after merge since you can only select known attributes of the index.

myitcv (approved):
LGTM

myitcv:
Closed via 72e7ae521ec241e0d09f24a9c7b42cf6ad738d74

PR #367, opened 2023-06-21
Hugo: Make anchor keyboard accessible

JorindeUsMedia:
- Add missing import to anchor scss file
- Make sure anchor icon also shows when using the keyboard to navigate the page
- Remove focus style on headings since headings are not focusable so this doesn't do anything

For: https://linear.app/usmedia/issue/CUE-242

### DIFF
--- a/hugo/assets/scss/components/anchor.scss
+++ b/hugo/assets/scss/components/anchor.scss
@@ -1,3 +1,4 @@
+@import '../config/sizes';
 @import '../mixins/screen';
 
 .anchor {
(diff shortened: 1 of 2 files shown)


myitcv:
What UI/UX changes should we be reviewing in this PR?

myitcv:
> What UI/UX changes should we be reviewing in this PR?

Ah, I see now.

JorindeUsMedia:
> > What UI/UX changes should we be reviewing in this PR?
> 
> Ah, I see now.

When you use your keyboard to browse docs pages you can see an outline for the anchor links but not the icon so it's not clear where you're focussing on. 

Screenshot from 2023-06-26 09-43-19

myitcv (approved):
LGTM - will import into GerritHub

PR #370, opened 2023-06-29
doc: replace ioutil references in Go/protobuf integration docs

nickfiggins:
Hi!

I was going through the tutorials / documentation to learn about cue and noticed the docs were using the deprecated ioutil package. This is a small update to use the `os` package's function instead.

### DIFF
--- a/content/en/docs/integrations/go.md
+++ b/content/en/docs/integrations/go.md
@@ -223,5 +223,5 @@ if err != nil {
     // handle error
 }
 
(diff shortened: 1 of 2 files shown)


myitcv:
The build failure can be ignored. Rather embarrassingly, the CI check that performs an early check for the `Signed-off-by` header is not case insensitive, and it should be. Will import into Gerrit now.

nickfiggins:
Oh awesome, sounds good! 


And whoops, let me know if it's easier if I amend the commit and I can do that instead 👍

PR #371, opened 2023-07-05
Update generate.md refer protobuf to golang

randomizedcoder:
Correct documentation to make it clear golang generation from protobuf definitions does not work.
Ref: https://github.com/cue-lang/cue/issues/581


myitcv:
@randomizedcoder apologies for dropping this. The content on the page you updated is vastly improved as part of the new site: https://cuelang.org/docs/concept/how-cue-works-with-protocol-buffers/

PR #372, opened 2023-07-07
Search: Improve searchable attributes

JorindeUsMedia:
- Remove category from search-filters
- Replace categories with tags in docs pages
- Rename category-overview and category-index to tags-overviews and tags-index
- Rename section to contentType
- Show contentType instead of categories in search-teaser.
- On pages of type 'docs' use the top level parent page as content type (language-guide, how-to etc)

For: https://linear.app/usmedia/issue/CUE-250

----
Testing: 
Searching itself for contentType can only be tested after merge because the algolia index gets updated when merging to alpha.
The json which will get uploaded to algolia you can see at: https://deploy-preview-372--cue.netlify.app/algolia.json

### DIFF
--- a/content/docs/howto/best-practices/en.md
+++ b/content/docs/howto/best-practices/en.md
@@ -2,6 +2,6 @@
 title: Best practices
 weight: 50
 draft: false
-layout: category-overview
-category: Best practices
+layout: tag-overview
+tag: Best practices
 ---
--- a/content/docs/howto/cue-command/en.md
+++ b/content/docs/howto/cue-command/en.md
@@ -2,6 +2,6 @@
 title: cue command
 weight: 40
 draft: false
-layout: category-overview
-category: cue command
+layout: tag-overview
+tag: cue command
 ---
--- a/content/docs/howto/ecosystem/en.md
+++ b/content/docs/howto/ecosystem/en.md
@@ -2,6 +2,6 @@
 title: Ecosystem
 weight: 60
 draft: false
-layout: category-overview
-category: Ecosystem
+layout: tag-overview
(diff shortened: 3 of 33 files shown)


JorindeUsMedia:
Noticed a bug where you can't search for facets with a space in it because we now split up the input using the space as a divider.. 
Have to think of something to handle spaces in tags & contenttypes.  Github uses double quotes if a value contains spaces: I will do something similiar.

JorindeUsMedia:
> Noticed a bug where you can't search for facets with a space in it because we now split up the input using the space as a divider.. Have to think of something to handle spaces in tags & contenttypes. Github uses double quotes if a value contains spaces: I will do something similiar. (https://github.com/cue-lang/cue/issues?q=is%3Aopen+is%3Aissue+label%3A%22v0.5+hangover%22+label%3AAccepted for example)

I made a separate issue for the space bug because it can be fixed separate from this pr and doesn't need to block it.

myitcv:
> Noticed a bug where you can't search for facets with a space in it because we now split up the input using the space as a divider.. Have to think of something to handle spaces in tags & contenttypes. Github uses double quotes if a value contains spaces: I will do something similiar. (https://github.com/cue-lang/cue/issues?q=is%3Aopen+is%3Aissue+label%3A%22v0.5+hangover%22+label%3AAccepted for example)

The GitHub approach sounds like the right answer to me.

myitcv (approved):
LGTM will import into Gerrit now.

myitcv:
Closed via 2cc25938b0125b533a4c01e02cd7c5559228f6c7

PR #373, opened 2023-07-07
Search: Handle spaces inside filters

JorindeUsMedia:
- Handle both values with and without quotes.
- Make sure spaces inside quotes are allowed but not with the non-quote version.

Now both these cases are working:
contentType:"Language Guide"  contentType:Language

For: https://linear.app/usmedia/issue/CUE-255


myitcv:
> The 2 regexes are a bit 'heavy'/

Measured how? 

I can't imagine that this is going to be particularly heavy weight given the load, so this approach looks fine for now. 

Will import into Gerrit.

myitcv:
Closed via dfc71d34f8c630e4b271fecf15829b15147922af

PR #374, opened 2023-07-19
Hugo: Search - Filters dropdowns & tags

JorindeUsMedia:
1. Add filters to search page for tags & content Type
- Add filter partial for each separate filter
- Style filters & filter bar
- Move algolia logo above bar to make space for filters
- Use html5 details & summary to handle toggling of filters and add js on top of html5-details to handle click on close button and close on click outside
- Add input field for fuzzy search + style input field
- Add fuzzy search to search in filterlist
- Render filter list via js
- Get tags from params toml so we're able to add color besides just the name
- Determine which filters are selected in list + add class is-selected
- Handle click on filters

2. Search teasers and tags
- Clean up list & teaser html & css so teasers look the same everywhere
- Add tags partial
- Adjust teaser to show tags and get the full tag from the tag config so we can we can use tag color in the html
- Handle click on tag in search teaser (search-results) via custom event

For: https://linear.app/usmedia/issue/CUE-245


myitcv (approved):
LGTM - will import into Gerrit now.

myitcv:
Closed via 9c596afcbc5525c96b4f64017433e91b08c0b453

PR #377, opened 2023-07-20
Hugo: add the meta description tag to the head of the html

ErikHendrikson:
This will add the meta tag with name="description" to the head of the html.

- the value of the tag will be set to the page's params description;

- with a fallback to the, by Hugo generated, summary of the page; and

- with a fallback to the general description of the Site Params;

For https://linear.app/usmedia/issue/CUE-252

### DIFF
--- a/hugo/layouts/partials/site/head.html
+++ b/hugo/layouts/partials/site/head.html
@@ -1,5 +1,9 @@
 <meta charset="utf-8">
 <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
+<meta name="description" content="{{ with .Description | $.RenderString | plainify }}{{ . -}}
+{{ else }}{{ with .Summary }}{{ replace ( trim ( . | plainify | htmlUnescape | chomp | truncate 160) " ")  "\n" "" -}}
+{{ else }}{{ with .Site.Params.description -}}{{- . | $.RenderString | plainify -}}
+{{ end }}{{ end }}{{ end -}}">
 {{ hugo.Generator }}
 {{ if eq (getenv "HUGO_ENV") "production" }}
 <meta name="robots" content="index, follow">


anne-usmedia (approved):

anne-usmedia (approved):

myitcv:
Thanks @ErikHendrikson - what happens in the case that a description is multi-line? Do things still work ok and generate valid HTML?

JorindeUsMedia (approved):

JorindeUsMedia:
> Thanks @ErikHendrikson - what happens in the case that a description is multi-line? Do things still work ok and generate valid HTML?

Is you use the description in the front-matter this should not contain empty lines and should not contain html tags etc. The fallback for hugo is Summary. See https://gohugo.io/content-management/summaries/  how hugo handles this. 
Note: Erik solution is the same as Hugo internals already uses for the open graph description tag.

JorindeUsMedia:
> > Thanks @ErikHendrikson - what happens in the case that a description is multi-line? Do things still work ok and generate valid HTML?
> 
> Is you use the description in the front-matter this should not contain empty lines and should not contain html tags etc. The fallback for hugo is Summary. See https://gohugo.io/content-management/summaries/ how hugo handles this. Note: Erik solution is the same as Hugo internals already uses for the open graph description tag.

Just noticed there is an open PR concerning empty lines etc in Summary.
Maybe we can use the solution proposed in the pr for meta-description. For the internal (opengraph, twitter etc) we have to either wait for hugo to merge this pr or handle them ourselves (not preferred) @ErikHendrikson 

https://github.com/gohugoio/hugo/issues/9818
https://github.com/gohugoio/hugo/pull/9856

JorindeUsMedia (changes requested):
See comment on pr. Maybe we can try to make this more solid.

hugo/layouts/partials/site/head.html
```diff
@@ -1,5 +1,6 @@
 <meta charset="utf-8">
 <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
+<meta name="description" content="{{ with .Description | $.RenderString | plainify }}{{ . }}{{ else }}{{ with .Summary }}{{ . | plainify | htmlUnescape | chomp | truncate 160 -}}{{ else }}{{ with .Site.Params.description }}{{ . | $.RenderString | plainify }}{{ end }}{{ end }}{{ end }}">
```
JorindeUsMedia, line 3: Maybe make this multiple lines? It's a bit hard to read and exceeds the max line length

ErikHendrikson: yes i agree

hugo/layouts/partials/site/head.html
```diff
@@ -1,5 +1,11 @@
 <meta charset="utf-8">
 <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
+<meta name="description" content="
+{{ with .Description | $.RenderString | plainify }}{{ . }}
```
JorindeUsMedia, line 4: you do need dashes for beginning and end tags now {{-   and -}} because now new lines inside the code also end up in the description tag

hugo/layouts/partials/site/head.html
```diff
@@ -1,5 +1,11 @@
...
+{{ else }}{{ with .Summary }}{{ . | plainify | htmlUnescape | chomp | truncate 160 -}}
+{{ else }}{{ with .Site.Params.description }}{{ . | $.RenderString | plainify }}
+{{ end }}{{ end }}{{ end }}
```
JorindeUsMedia, line 7: is it correct we have 3 end tags here? I though and if else only needed 1? (not sure though..)

JorindeUsMedia (changes requested):

hugo/layouts/partials/site/head.html
ErikHendrikson: yes, its needed. https://github.com/gohugoio/hugo/issues/9818 (example)

JorindeUsMedia (approved):

jpluscplusm:
Imported into Gerrit

jpluscplusm:
Closed via 0d3a50bdf073a69165b2c50b40449d6b27a79dc2

PR #378, opened 2023-07-24
Hugo: Build script improvements

JorindeUsMedia:
- Add minify flag to hugo build in order to have minified html output which results in a better speed performance.
- Remove npm run icons task from the build script since it's not needed because the icon sprites itself are in the asset folder in the src. This script only needs to run manually when an icon is added (see hugo/readme.md). Running this in the pipeline can even cause git issues because this way the build script might update the sprite asset in the src folder which causes a git diff.

For: https://linear.app/usmedia/issue/CUE-253

### DIFF
--- a/_scripts/build.bash
+++ b/_scripts/build.bash
@@ -10,10 +10,12 @@ cd "$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )/.."
 skipcache=""
 norun="true"
 time=""
+minify="--minify"
 if [ "${NETLIFY:-}" != "true" ]
 then
 	time="time -p"
 	norun="false"
+	minify=""
 	if [ "${CI:-}" == "true" ]
 	then
 		skipcache="--skipcache=true"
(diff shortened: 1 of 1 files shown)


_scripts/build.bash
```diff
@@ -29,5 +29,4 @@ bash _scripts/runPreprocessor.bash execute --debug=all --norun=$norun $skipcache
 # Main site
 cd hugo
 $time npm ci
-$time npm run icons
-$time hugo --cleanDestinationDir $@
+$time hugo --cleanDestinationDir --minify $@
```
myitcv, line 32: I'd suggest setting a variable for this above (like `time`, empty by default), such that minification only happens as part of the Netlify run. Then use that variable as the argument.

myitcv (approved):
LGTM with one suggestion.

myitcv:
@jpluscplusm is checking this out locally as a means of getting up to speed, but still LGTM!

jpluscplusm (approved):
LGTM from a non-FE-dev perspective! I've run it up locally, and I correctly /don't/ see minified HTML locally, but do see it on the preview app. It renders identically on the couple of pages I checked.

jpluscplusm:
Closed via 7efa2818e5d04001ca58678be12453db6f26dd74

PR #382, opened 2023-07-28
hugo: blog overview

ErikHendrikson:
This will add the blog overview page.

- four example pages have been added: /blog/example, blog/example2,
blog/example3, blog/example4;

- the teaser--blog template has been added;

- the list and list--overview BEM-classes are used to render
the several blogs in the blogs/list.html template;

- the first teaser--blog gets a modifier "featured", which will be used
to give extra styling for the featured teaser and the path of the directory is
passed on through 'dict';

- the featured teaser-blog gets, starting at tablet device
a bigger heading style, an excerpt of the content with grey color,
and the date is not displayed;

- the transition property has been set on the teaser__link instead
of the teaser class itself;

- the teaser--blog image has a standard height of 130px, from tablet
this will be 340px for the featured teaser--blog and 164px for the rest;

- the teaser template and the teaser--blog template have been moved
from the _default to the partials directory;

- blog has been added to the navigation menu in the header;

- blog has been added to the footer menu in the footer;

For https://linear.app/usmedia/issue/CUE-258

Signed-off-by: Erik Hendrikson <erik.hendrikson@usmedia.nl>


myitcv:
There is lots of debug logging noise in the failed build output here, which will be fixed by https://review.gerrithub.io/c/cue-lang/cuelang.org/+/557038.

However, if you look for the `** ` prefix to a line, you will see the error comes from:

```
...
** /home/myitcv/dev/cuelang/cuelang.org/content/blog/_en.html: failed to find end of header
...
```

If you look at the file, it appears to be fine. The problem is that it is missing a trailing newline. 

This appears to be an option that you can turn on with VSCode:

https://hello-sunil.in/insert-newline-vs-code/

hugo/assets/scss/components/blogs.scss
```diff
@@ -0,0 +1,92 @@
+@import '../config/colors';
+@import '../config/sizes';
+@import '../mixins/container';
+@import '../mixins/list-reset';
+@import '../mixins/stretch';
+
+.blogs {
```
JorindeUsMedia, line 7: we already have a list bem component set up for this. list--overview (from an early design). You can use that bem and move the styles there.

hugo/assets/scss/components/blogs.scss
```diff
@@ -0,0 +1,92 @@
...
+            .teaser {
+                .media {
+                    height: 130px;
+                }    
+            }
+
+            &:first-child {    
+                grid-column: 1 / -1;
+
+                .teaser {
```
JorindeUsMedia, line 32: teaser shouldn't be styles by the list itself. You can add a modifiier to teaser which handles the featured teaser styles (add a class in the blog list when index = 1 (or 0 not sure hugo range is 0 indexed) teaser--featured or something? or teaser--highlight

hugo/assets/scss/components/teaser.scss
```diff
@@ -65,6 +65,42 @@
...
     }
 
+    &--blog {
+        #{ $self }__content {
+            display: flex;
+            flex-direction: column;
+            gap: 14px;
+        }   
+
+        #{ $self }__meta, #{ $self }__excerpt, #{ $self }__title {
```
JorindeUsMedia, line 75: please move each selector to a new line

hugo/assets/scss/components/teaser.scss
```diff
@@ -65,6 +65,42 @@
...
+            margin: 0;
+        } 
+
+        #{ $self }__title {
+            @include style-heading-3;
+        }
+
+        .media {
```
JorindeUsMedia, line 83: is i tcorrect we have the image class as teaser__image but the media class just as media?

hugo/layouts/partials/teaser--blog.html
```diff
@@ -0,0 +1,29 @@
+<article class="teaser teaser--blog">
+    {{ if .Params.Image }}
+        <div class="media">
```
JorindeUsMedia, line 3: is i tcorrect we have the image class as teaser__image but the media class just as media? Also do we need both a div classs media and figure class teaser__image. ?

hugo/layouts/partials/teaser--blog.html
```diff
@@ -0,0 +1,29 @@
...
+            {{ end }}
+        </div>
+        <p class="teaser__meta">{{ .Date.Format ($.Param "time_format") }}</p>
+        <p class="teaser__excerpt">{{ .Summary | truncate 195 }}</p>
+        <div class="link teaser__readmore">
+            <span class="link__text">{{ T "ui_read_more" }}</span>
+        </div>
+    
+        <a class="teaser__link" aria-label="{{ T "ui_read_more" }} - {{ .LinkTitle }}" href="{{ .RelPermalink }}">
+            <span>{{ T "ui_read_more" }}</span>
```
JorindeUsMedia, line 25: aria-label and text itself are now the same. Won't that be a problem for screenreaders where the same text gets to be read twice?

JorindeUsMedia (changes requested):

hugo/assets/scss/components/section.scss
```diff
@@ -62,6 +62,12 @@
         }
     }
 
+    &--blog {
```
JorindeUsMedia, line 65: we can just use the regular section title, ignore the design here (old section title styling was used there)

JorindeUsMedia (changes requested):
Hover state of the teaser can be nicer. Maybe don't do a backgrond color but zoom the image in a bit? 

Also check mobile: we can remove padding on teaser on mobile so the left side of the teaser alligns with the section title, nav etc (so no double padding as we have now)

JorindeUsMedia:
Forgot about this: For all images we use we need permission so please one we already use in examples pages (carousel, image)

hugo/assets/scss/components/teaser.scss
```diff
@@ -55,19 +54,102 @@
...
-        &:hover {
-            background: transparentize($c-blue, 0.94);
-        }
-
         &:focus-visible {
             border-color: var(--teaser-border-focus, $c-blue);
         }
     }
 
+    &:not(#{ $self }--blog) {
```
JorindeUsMedia, line 62: the not selector is very specific so I want to avoid using it because it's almost impossible to overrule this line now. Can't we leave the default hover as is and overrule it withing the blog modifier?

JorindeUsMedia:
Not sure if we already want to have blogs overview link in the menu (it's filled with examples pages now) or if we want to wait for that for the real blogs to land. @myitcv  what do you think?

hugo/assets/scss/components/teaser.scss
ErikHendrikson: yes, agreed

ErikHendrikson: there was also a transition missing on the teaser__link, therefore not working, so I put the transition on teaser__link instead of the .teaser, so this is also fixed now.

myitcv:
> Not sure if we already want to have blogs overview link in the menu (it's filled with examples pages now) or if we want to wait for that for the real blogs to land. @myitcv what do you think?

Adding to the link menu is fine for now, thanks.

hugo/assets/scss/components/teaser.scss
```diff
@@ -29,45 +27,129 @@
...
         color: $c-blue--dark;
         font-size: 1.5rem;
-        margin: 0 0 0.4rem;
+        margin: 0 0 .4rem;
     }
 
     &__meta {
         color: var(--teaser-meta-color, $c-grey--dark);
-        font-size: 0.8125rem;
+        font-size: .8125rem;
```
JorindeUsMedia, line 35: please use 0. notation 
we can't force this in editrconfig but prettier etc should do this automatically

hugo/assets/scss/components/teaser.scss
```diff
@@ -11,9 +11,7 @@
     background-color: var(--teaser-background, transparent);
     color: var(--text-color);
     height: 100%;
-    padding: $p-gutter;
```
JorindeUsMedia, line 14: removing this breaks the normal teasers on mobile

hugo/assets/scss/components/teaser.scss
```diff
@@ -11,9 +11,7 @@
...
     position: relative;
-    transition: background-color 0.2s;
```
JorindeUsMedia, line 16: this is needed for the normal teaser which does have a background change on focus

hugo/assets/scss/components/teaser.scss
```diff
@@ -29,45 +27,129 @@
...
 
         span {
             @include sr-only;
         }
 
         &:focus,
         &:hover {
-            background: transparentize($c-blue, 0.94);
+            background: transparentize($c-blue, .94);
+            transition: background-color .2s;
```
JorindeUsMedia, line 59: putting the transition on focus itself means it only transitions one way around, so when focus > no focus and not the other way around anymore

JorindeUsMedia (changes requested):

hugo/assets/scss/components/teaser.scss
ErikHendrikson: There doesn't seem to be a background-color change on focus in .teaser though

JorindeUsMedia (approved):

JorindeUsMedia:
> We appear to have lost the date a blog was posted at, compared to the current alpha. Compare:
> 
> https://alpha.cuelang.org/blog/example/
> 
> with the new:
> 
> https://deploy-preview-382--cue.netlify.app/blog/example/
> 
> Is that intentional?
> 
> If so, will the author, date, tags etc get added in a later PR?

Yes both authors and tags will be picked up separately. For tags there is a draft PR open. For authors we're waiting for a new design.. 
So we have split up blog and overall detal pages stuf into: blog template, blog overview (teasers), date& reading time, tags & authors

myitcv (approved):
LGTM - will import into Gerrit now.

hugo/layouts/partials/teaser.html
```diff
@@ -9,6 +9,7 @@ <h2 class="teaser__title">{{ .Title }}</h2>
     <p class="teaser__meta">{{ .Date.Format ($.Param "time_format") }}</p>
     <p class="teaser__excerpt">{{ .Summary | truncate 300 }}</p>
     <div class="link teaser__readmore">
+        hoi
```
JorindeUsMedia, line 12: There's still some test text here

JorindeUsMedia (changes requested):

myitcv:
Closed via 2759088ce30015f3767765640cfbc10ad4602220

hugo/layouts/partials/teaser.html
myitcv: This will need to be addressed in a follow-up

JorindeUsMedia: Made a linear issue: https://linear.app/usmedia/issue/CUE-277

PR #383, opened 2023-08-01
hugo: detail/content page header meta

ErikHendrikson:
This will add meta details to the article header of the single pages.

- this will apply to default single pages, blog single pages and docs
single pages;

- in these single page templates the $meta variable is constructed, being
the .Date of the content, the .Params.meta of the content, a combination
of both or none. If there is meta data, then the meta partial will be
rendered;

- in the meta partial each meta item, if provided with .type and .value,
will be rendered in a list.

- if the type of the meta is readingTime, the translated text of ui_min_read
will be put after the value;

- the link-type has been removed;

- in meta.scss the bullet-separator between two meta items is applied;

- the example page of the front-matter, '/examples/front-matter', has been
updated;

- the changes are visible beneath the header on default single pages,
blog single pages and docs single pages;

For https://linear.app/usmedia/issue/CUE-262

### DIFF
--- a/content/examples/basic/front-matter/en.md
+++ b/content/examples/basic/front-matter/en.md
@@ -1,6 +1,10 @@
 ---
 title: Front Matter
 weight: 14
+date: "2023-08-01"
+meta:
+    - type: readingTime
+      value: 5
 tags:
     - Use encodings in CUE
     - Language
@@ -37,18 +41,20 @@ math
 : You can add `math: true` to the front matter, if you want to enable [MathJax](https://www.mathjax.org/) on a content page.
 
 meta
-: You can show additional meta information below the title of a content page. This is map of items consisting of a type and value (and optionally a link).
+: You can show additional meta information below the title of a content page. This is a map of items consisting of a type and value.
+: If you include the type readingTime, 'min Read' will be added to the given value.
+
+: When you include a date in the front-matter, it will be automatically included in the meta information.
+
+: Available types:
+
+- readingTime (value is in minutes)
+
 ```
 meta:
-- type: link
-  value: Website
-  link: https://www.usmedia.nl
-- type: person
-  value: Joop Verkerke
-- type: other
-  value: Some info
+- type: readingTime
+  value: 5
 ```
-The type will be mapped to an icon, see `/assets/svg/ui` for a full list of possible types.
 
 notification
 : You can add a sticky notification on the current page. The `content` allows for simple markdown options.
--- a/hugo/layouts/partials/meta.html
+++ b/hugo/layouts/partials/meta.html
@@ -1,41 +1,17 @@
-{{/*
-    items is map with
-      - type *
-      - value
-      - link
-
-    * Possible types (for more, look in /asssets/svg/ui)
-      - office
-      - person
-      - link
-      - date
-      - time
-      - location
-      - other (== mapped to cue)
-*/}}
 {{ $items := . }}
 
 {{ if $items }}
     <ul class="meta">
-        {{ range $items }}
-            <li class="meta__item">
-                {{ if .type -}}
-                    <div class="meta__icon">
-                        {{- if eq .type "other" -}}
-                            {{- partial "icon.html" (dict "icon" "cue") -}}
-                        {{- else -}}
-                            {{- partial "icon.html" (dict "icon" .type) -}}
-                        {{- end -}}
-                    </div>
-                {{- end -}}
-                <div class="meta__value">
-                    {{- if .link -}}
-                        <a class="meta__link" href="{{ .link }}"><span>{{ .value }}</span></a>
-                    {{- else -}}
-                        <span>{{ .value }}</span>
-                    {{- end -}}
-                </div>
-            </li>
+        {{ range $index, $item := $items}}
+            {{ if and .type .value }}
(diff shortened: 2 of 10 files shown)


hugo/layouts/partials/meta.html
```diff
@@ -4,7 +4,7 @@
       - value
       - link
 
-    * Possible types (for more, look in /asssets/svg/ui)
+    * Possible types (for more, look in /assets/svg/ui)
```
JorindeUsMedia, line 7: icons are removed from the partial so these instructions can also be removed

content/community/example/en.md
JorindeUsMedia, line 1: I want to avoid to have this many example pages since we already have real content for community and don't want to clutter it with example pages just for testing.. Can you create an example page for header/meta etc within the examples sections? So an example content page within the examples section and this example pages has both date & reading time set so we can use that for testing.

JorindeUsMedia (changes requested):

JorindeUsMedia (commented):
Code looks fine but just a small visual comment; The dot between the meta items is not in the middle: there seems to be more space on the left side than on the right.

JorindeUsMedia (changes requested):

JorindeUsMedia (approved):

myitcv:
Looks like this needs a rebase, please!

content/examples/basic/front-matter/en.md
```diff
@@ -1,6 +1,13 @@
 ---
 title: Front Matter
 weight: 14
+date: "2023-08-01"
+meta: 
+- type: readingTime
+  value: 1
+- type: link
```
JorindeUsMedia, line 8: discussed with Peter: we don't want to have a back link here because there already is a back link before the title so it's not needed and with all the other header items (authors, tags, etc) it will be to crowded

JorindeUsMedia (changes requested):

content/examples/basic/front-matter/en.md
ErikHendrikson: This can be a link to anything though... still want to remove it?

JorindeUsMedia: Yes I think so because it isn't in the designs.

JorindeUsMedia (approved):

content/examples/basic/front-matter/en.md
```diff
@@ -1,6 +1,10 @@
 ---
 title: Front Matter
 weight: 14
+date: "2023-08-01"
+meta:
```
jpluscplusm, line 5: I'd feel happier reviewing this with a couple of test examples on this page that /weren't/ the special-cased `readingTime` metadata field or the top-level `date` field.

Please could you add a couple of test metadata fields that show off the rendering capabilities (or limits we need to be aware of) when using this feature?

```suggestion
meta: 
- type: Test Type 1!
  value: A test value
- type: test `type` 2?
  value: Another test value
```

jpluscplusm (changes requested):

content/examples/basic/front-matter/en.md
jpluscplusm: (examples are deliberately in reverse sort order, so any sorting imposed by the code is visible)

JorindeUsMedia: for now we only have date and the meta item ReadingTime. In an earlier design we had more types so Erik adjusted the existing html partial and removed the types we don't have but we kept the partial in case we want to add more meta later.
We can list the available types we have now in the example page which for now is just Reading Time.

Peter is also working on the redesign of the whole content header which he will finish today and we want to start working on that in the afternoon. In the new header design everything comes to together: authors, meta (date + reading time), tags. So the styling of the meta Erik made for this PR is a start but will be finalized when we have the new header design in a separate story and pr

JorindeUsMedia: @jpluscplusm I've made the example more clear that for now we only have type 'readingTime' for meta.

jpluscplusm: @JorindeUsMedia My apologies - I'd misinterpreted https://github.com/cue-lang/cuelang.org/pull/383/files#diff-f9a6e91a7670ffb79caf08157f4d7619ed95c4c74f4e89994bdaa0f1e3c58bb0R5 as being an override for `readingTime`, but which caused any /other/ `meta.` type+value to be rendered unchanged. Thanks for the correction!

jpluscplusm (approved):

content/examples/basic/front-matter/en.md
JorindeUsMedia: @jpluscplusm I understand the confusion :) The old meta code from the earlier design was not in use anymore so for readingTime we took the old code and adjusted it a bit.

jpluscplusm:
This has a merge conflict, now that https://github.com/cue-lang/cuelang.org/pull/384 has been merged. Back over to USM folks ...

JorindeUsMedia:
> This has a merge conflict, now that #384 has been merged. Back over to USM folks ...

I will fix the conflict :) I'll let you know when it's done

jpluscplusm:
Closed via da1a35374e4d98a0ac66fb4ebd34bf1e7637cee8

PR #384, opened 2023-08-04
hugo: content/detail pages - show tags on page

ErikHendrikson:
This will add tags to the single pages

- in the "tags" partial, the tags from the front-matter of the page
are rendered;

- the front-matter example page has an added section of tags, and tags are
added to it's own front-matter;

- the tags are shown in the article-header on a single page, normal sized,
and in the article-footer on a single page, large sized;

- a new color has been added, c-lavender--light, to highlight tags that
are hovered over of focused on:

- the article--header partial has been created, and is rendered on all single
pages;

- the article meta and tags are put inside article__info, therefore
the margin styling has been moved from the .meta class to the .article__info
class;

- if there are tags, meta-data, or a date provided on the page front-matter,
the article header will receive a modifier "article__header--extended", which
will modify the margins in the header title, the header info and the header
itself;

- the article footer is now styled with a separator, so the separator
is removed from single pages and list pages where it was applied;

- the separator modifier has been removed from separator.scss, since
the article__footer is now styled with a separator;

For https://linear.app/usmedia/issue/CUE-236

### DIFF
--- a/content/examples/basic/front-matter/en.md
+++ b/content/examples/basic/front-matter/en.md
@@ -1,6 +1,10 @@
 ---
 title: Front Matter
 weight: 14
+tags:
+    - Use encodings in CUE
+    - Language
+    - Ecosystem
 ---
 
 Hugo uses front-matter to add meta information to a content page. This can be done various formats - in the CUE website we prefer **YAML** (To be decided).
@@ -63,3 +67,11 @@ notification:
         icon: 'download'
 ```
 You can also add a site-wide notification via the config file for params (ex. `config/_default/params.toml`)
+
+tags
+: By adding tags in the front-matter, they will be displayed in both the header and footer of the content page. Tags link to `/search?q=tag:{name}` and initiate a search for all documents that contain the respective tag.
+```
+tags:
+- Ecosystem
+- Language
+```
--- a/hugo/assets/scss/components/article.scss
+++ b/hugo/assets/scss/components/article.scss
@@ -22,9 +22,35 @@
         position: relative;
     }
 
+    &__footer {
+        border-top: solid 2px $c-yellow;
+        display: flex;
+        flex-direction: column;
+        gap: 2rem;
+        margin: 0 auto;
+        opacity: 1;
+        padding: 1.5rem 0;
+    }
+
+    &__info {
+        column-gap: 1rem;
+        display: flex;
+        flex-wrap: wrap;
+        margin: 0.5rem 0 2rem;
+        row-gap: 0.75rem;
+    }
+
     &__header {
         position: relative;
         width: 100%;
+
+        &--extended {
+            margin: 2.5rem 0 2rem;
+    
+            #{ $self}__title {
+                margin: 0;
+            }
+        }
     }
 
     &__title {
--- a/hugo/assets/scss/components/list.scss
+++ b/hugo/assets/scss/components/list.scss
@@ -36,7 +36,7 @@
     &--tags {
         display: flex;
         flex-wrap: wrap;
-        gap: 4px;
+        gap: 7px;
     }
 
     @include screen($screen-simple) {
@@ -51,6 +51,10 @@
 
     @include screen($screen-normal) {
         gap: 2rem;
+
+        &--tags {
+            gap: 7px;
+        }
 
         &--overview {
             grid-template-columns: repeat(3, 1fr);
--- a/hugo/assets/scss/components/tag.scss
+++ b/hugo/assets/scss/components/tag.scss
@@ -2,15 +2,18 @@
 @import '../config/typography';
 
 .tag {
+    --tag-background: #{ $c-blue--lighter };
+
     align-items: center;
-    background: var(--tag-background, transparentize($c-white, 0.5));
-    border-radius: 24px;
+    background: var(--tag-background);
+    border-radius: 1.5rem;
+    color: $c-black;
     display: flex;
     font-size: 0.8125rem;
     font-weight: $weight-medium;
     gap: 0.5rem;
-    height: 22px;
-    line-height: 22px;
+    line-height: 1.25;
+    min-height: 1.5rem;
     padding: 0 0.6rem 0 0.5rem;
     position: relative;
     text-decoration: none;
@@ -19,7 +22,7 @@
 
     &:focus-visible,
     &:hover {
-        --tag-background: #{ $c-white };
+        --tag-background: #{ $c-lavender--light };
     }
 
     &::before {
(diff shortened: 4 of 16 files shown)


hugo/layouts/partials/tags.html
```diff
@@ -1,15 +1,20 @@
-{{ $tags:= .tags| default slice }}
+{{ $context := .context | default slice }}
+{{ $tags := $context.Params.tags | default slice }}
+{{ $tagOptions := $context.Site.Params.Tags | default slice }}
+{{ $filteredTags := where $tagOptions "name" "in" $tags | default slice }}
```
JorindeUsMedia, line 4: I would make it so the tags partial doesn't have to do filtering but gets the full tags allready. This way the partial is better reusable in all cases. How it gets the tags logic should not be inside the partial in my opinion

hugo/assets/scss/components/tag.scss
```diff
@@ -19,7 +22,9 @@
 
     &:focus-visible,
     &:hover {
-        --tag-background: #{ $c-white };
+        --tag-background: #DDE3FD;
+        
+        transform: scale(1.03);
```
JorindeUsMedia, line 27: the hover is a bit heavy now and transition is missing which is making it a 'clunky' , especially when there's going to be multiple tags.. Maybe just change the background color? 
Also put the hex variable in the colors.scss

hugo/assets/scss/components/tag.scss
```diff
@@ -31,6 +36,11 @@
         width: 12px;
     }
 
+    &--large {
+        min-height: 44px;
+        padding: 0 21px 0 18px;
```
JorindeUsMedia, line 41: is there a reason we use pixels here instead of rem?

hugo/layouts/_default/single.html
```diff
@@ -3,18 +3,17 @@
...
-                <div class="article__meta">{{ partial "meta" . }}</div>
-                {{- end }}
-            </header>
+            {{- partial "article--header" . -}}
             <div class="article__content">
                 {{- .Content -}}
             </div>
+            <footer class="article__footer">
+                {{- if .Params.tags -}}
+                    {{- partial "tags.html" (dict "modifier" "large" "context" . ) -}}
```
JorindeUsMedia, line 12: don't pass the full context to tags: partials should not be dependent on whatever context they are in but should be usable anywhere

hugo/assets/scss/components/article.scss
```diff
@@ -22,6 +22,24 @@
...
+        margin: 0 auto;
+        opacity: 1;
+        padding: 24px 0;
+    }
+
+    &__info {
+        column-gap: 1rem;
+        display: flex;
+        flex-wrap: wrap;
+        margin: 1.5rem 0;
```
JorindeUsMedia, line 39: The margin is too big. The heading should be compact, especially since there's a lot of info there (date, reading time, title, authors, tags etc). Also visual the distance to the title is now too big which makes the tags 'float' (not belong anywhere..). Maybe use 0.5 rem?
Screenshot from 2023-08-07 10-28-50

hugo/assets/scss/components/list.scss
```diff
@@ -48,6 +48,10 @@
     @include screen($screen-normal) {
         gap: 2rem;
 
+        &--tags {
+            gap: 7px;
```
JorindeUsMedia, line 52: both regular and screen normal are now 7px so this line can go

hugo/assets/scss/components/tag.scss
```diff
@@ -2,15 +2,18 @@
...
+
     align-items: center;
-    background: var(--tag-background, transparentize($c-white, 0.5));
+    background: var(--tag-background);
     border-radius: 24px;
+    color: $c-black;
     display: flex;
     font-size: 0.8125rem;
-    font-weight: $weight-medium;
+    font-weight: $weight-semibold;
```
JorindeUsMedia, line 13: In design medium is used, not semibold

JorindeUsMedia (changes requested):

content/examples/basic/front-matter/en.md
```diff
@@ -63,3 +67,11 @@ notification:
         icon: 'download'
 ```
 You can also add a site-wide notification via the config file for params (ex. `config/_default/params.toml`)
+
+tags
+: By adding tags in the front-matter, they will be displayed in both the header and footer of the content page. Tags link to `/search?q=tag:{name}` and initiate a search for all documents that contain the respective tag.
+```
+tags:
```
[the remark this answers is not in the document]
ErikHendrikson: Do you mean `/search?q=tag`? Because it does link to the specified tag-name?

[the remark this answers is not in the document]
JorindeUsMedia: > Do you mean `/search?q=tag`? Because it does link to the specified tag-name?

No, check for example the docs where tags are already in use.
For instance:

tags:
    - Ecosystem
    - cue command

hugo/assets/scss/components/tag.scss
ErikHendrikson: no, but I will use rem from now on!

hugo/assets/scss/components/article.scss
```diff
@@ -22,9 +22,35 @@
...
+        flex-wrap: wrap;
+        margin: 0.5rem 0 2rem;
+        row-gap: 0.75rem;
+    }
+
     &__header {
         position: relative;
         width: 100%;
+
+        &--extended {
```
JorindeUsMedia, line 47: I would maybe solve this differently later when we have a lot of things in the header but lets pick that up when we do the whole header redesign

JorindeUsMedia (approved):

jpluscplusm:
@ErikHendrikson Please could you mark completed conversations that you consider "dealt with" as "Resolved", so I can definitively tell that there's nothing /unresolved/ that might result in more pushes to this branch? Ta!

JorindeUsMedia:
> @ErikHendrikson Please could you mark completed conversations that you consider "dealt with" as "Resolved", so I can definitively tell that there's nothing /unresolved/ that might result in more pushes to this branch? Ta!

Erik is off today and I can't mark them as resolved but I checked all my comments and all of them are resolved..

jpluscplusm:
@JorindeUsMedia Thank you!

jpluscplusm (approved):
The tag header and footer (e.g. on https://deploy-preview-384--cue.netlify.app/examples/basic/front-matter/) looks /great/!

jpluscplusm:
imported as https://review.gerrithub.io/c/cue-lang/cuelang.org/+/1167557

jpluscplusm:
Closed via dd3a2d7d7955089f8af4b5e7a290a571d71254e3

PR #391, opened 2023-08-15
Hugo: search should do OR of content-type

ErikHendrikson:
This will modify the Algolia search filter. When you select multiple content-types from the dropdown, the filtering will be based on any of the chosen types. Consequently, the filtering will not hinge on the requirement that all selected content types are present simultaneously.

For https://linear.app/usmedia/issue/CUE-267

### DIFF
--- a/hugo/assets/ts/helpers/search.ts
+++ b/hugo/assets/ts/helpers/search.ts
@@ -33,9 +33,14 @@ export const getInputNameForFacet = (facet: SearchFacet): SearchInputFacetName =
 export const mapToAlgoliaFilters = (tagsByFacet: SearchFacets, operator: FilterOperator = FilterOperator.AND): string => {
     return Object.keys(tagsByFacet)
         .map((facet) => {
+            const facetOperator = {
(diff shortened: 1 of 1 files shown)


JorindeUsMedia (approved):

jpluscplusm (approved):
I don't believe this is functionally reviewable until after its integration, as only the tip of the alpha branch gets an Algolia search backend.

jpluscplusm:
Imported as https://review.gerrithub.io/c/cue-lang/cuelang.org/+/1167576

JorindeUsMedia:
> 

This change is on the client side search to algolia so not the indexing (backend) part. 
You can test it on the search page. For instance: https://deploy-preview-391--cue.netlify.app/search/?q=cue%20contentType:Introduction%20contentType:%22Language%20Guide%22
This now has results from both contentType Language Guide and Introduction. 
Before this change it had no results because it tried to find pages with both Language Guide and Introduction as the content type which can never happen.

jpluscplusm:
Closed via 13d3b5e9630d49017b0b1fddddca01ed287e4d47

PR #392, opened 2023-08-15
Hugo: improve distinguishability of inline code

anne-usmedia:
- Added background + border as seen in design
- Only added this when the inline code is inside a p. This because previous wishes to remove the styling from titles

For https://linear.app/usmedia/issue/CUE-279

### DIFF
--- a/hugo/assets/scss/base/base.scss
+++ b/hugo/assets/scss/base/base.scss
@@ -339,6 +339,15 @@ code {
     font-style: normal;
     font-weight: inherit;
     line-height: inherit;
+
+    p & {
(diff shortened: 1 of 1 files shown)


JorindeUsMedia (approved):
I've checked on https://deploy-preview-392--cue.netlify.app/examples/basic/code-block/ and it seems to be fine. The grey is a bit light but it's the same grey as the bigger code blocks so I think that's fine

myitcv (approved):
LGTM. Will import now.

jpluscplusm:
Closed via 0e229156965350ab5437ef9061452a23218009c9

PR #393, opened 2023-08-16
Internal/Hugo: Update Hugo to 0.117.0

anne-usmedia:
- Update Hugo to 0.117.0
- Change needed Hugo version in readme
- Fix deprecated warnings for - languages.toml - config.toml
- Fix warning for .File.UniqueID: added a `with` for the .File

For https://linear.app/usmedia/issue/CUE-243

### DIFF
--- a/hugo/layouts/_default/list.algolia.json
+++ b/hugo/layouts/_default/list.algolia.json
@@ -18,8 +18,12 @@
     {{ $cutoff := 1000 -}}
     {{ $index := .index -}}
     {{ $skipWords := mul $cutoff $index }}
+    {{ $fileID := replace $page.Title " " "_"|lower }}
+    {{- with $page.File -}}
+        {{- $fileID = .UniqueID -}}
+    {{- end -}}
 
-    {{- $objectID := (print $page.File.UniqueID "_" (add $index 1)) -}}
+    {{- $objectID := (print $fileID "_" (add $index 1)) -}}
     {{- $title :=  $page.Title -}}
     {{- $link := $page.RelPermalink -}}
     {{- $publishDate := $page.PublishDate -}}
--- a/.github/workflows/trybot.yml
+++ b/.github/workflows/trybot.yml
@@ -167,7 +167,7 @@ jobs:
         name: Install Hugo (${{ runner.os }})
         uses: peaceiris/actions-hugo@v2
         with:
-          hugo-version: 0.115.4
+          hugo-version: 0.117.0
           extended: true
       - if: runner.os == 'macOS'
         name: Install Hugo (${{ runner.os }})
--- a/.github/workflows/update_tip.yml
+++ b/.github/workflows/update_tip.yml
@@ -79,7 +79,7 @@ jobs:
         name: Install Hugo (${{ runner.os }})
         uses: peaceiris/actions-hugo@v2
         with:
-          hugo-version: 0.115.4
+          hugo-version: 0.117.0
           extended: true
       - id: go-mod-cache-dir
         name: Get go mod cache directory
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ issue a `cuelang.org: ` prefix.
 
 * [NodeJS](https://nodejs.org/) `v18.x`
 * [Go](https://golang.org/dl/) `>= go1.20` (not needed for only running `hugo`)
-* [Hugo](https://github.com/gohugoio/hugo/releases) `>= v0.111.3`
+* [Hugo](https://github.com/gohugoio/hugo/releases) `>= v0.117.0`
   (["extended"](https://gohugo.io/troubleshooting/faq/#i-get-this-feature-is-not-available-in-your-current-hugo-version)
   binary version)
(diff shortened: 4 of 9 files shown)


hugo/layouts/_default/list.algolia.json
```diff
@@ -19,53 +19,55 @@
...
-    {{- $objectID := (print $page.File.UniqueID "_" (add $index 1)) -}}
-    {{- $title :=  $page.Title -}}
-    {{- $link := $page.RelPermalink -}}
-    {{- $publishDate := $page.PublishDate -}}
-    {{- $summary := $page.Summary -}}
-    {{- $content := "" -}}
-    {{- $tags := cond (isset $page.Params "tags") ($page.Params.tags) "" -}}
-    {{- $contentType := $page.FirstSection.Title -}}
+    {{- with $page.File -}}
+        {{- $objectID := (print $page.File.UniqueID "_" (add $index 1)) -}}
```
JorindeUsMedia, line 23: would suggest to move this logic wiithin objectId
something like this
{{- $fileID := $page.File ? $page.File.UniqueID  : 'page' -}} > this is not valid code but just to give an idea
{{- $objectID :=  (print fileID "_" (add $index 1)) -}}

hugo/layouts/_default/list.algolia.json
```diff
@@ -18,8 +18,12 @@
     {{ $cutoff := 1000 -}}
     {{ $index := .index -}}
     {{ $skipWords := mul $cutoff $index }}
+    {{ $fileID := delimit (shuffle (split (md5 $page.Title) "" )) "" }} {{/* generates a "random" string based on the page title */}}
```
JorindeUsMedia, line 21: maybe movie this to line 23-24 in the else statement because reading it now looks a bit like the random id is the default (which we do now want of course because that would result in a lot of duplicate entries in algolia)

JorindeUsMedia: another note: don't use shuffle on the title because that would result in a different id with each render. Algolia uses objectId to check if an entry exists or not and generating random as default will cause duplicates in algolia because the id changes

myitcv: What is the uniqueness we are trying to go for here? Unique per page? Or unique per page per commit?

JorindeUsMedia: in 99.99% of the cases we do have file-Id which is unique but we have to check for this because it can be empty and right now Hugo gives a warning on build for this line. That's why we had to have a fallback

myitcv: In what cases can the file-id be empty? 

Is there anything we can/should rely on as far as the preprocessor is concerned?

JorindeUsMedia: the warning in the build was:
debug: hugo: WARN  .File.UniqueID on zero object. Wrap it in if or with: {{ with .File }}{{ .UniqueID }}{{ end }}
@anne-usmedia Did you find out which page triggered this error?

anne-usmedia: The tags (/tags/language) pages are the ones that do not have a .File, and therefore wouldn't have an objectID in the algolia.json.

JorindeUsMedia: I've created a new issue for the tag-overview pages which should not exist: https://linear.app/usmedia/issue/CUE-310
But this is already happening in current alpha so isn't caused by the update to hugo 0.117.
We still should have a fallback for non file pages since hugo gives warnings about them now but we have to make sure they don't change on build because they need to be the same as algolia uses them to identify the item. So we can't use randomize or loop index because these change Maybe use slugified page title or something as a fallback for automaticly genertated pages? Even if we do not use them we do need to have something in place because we want to get rid of the current warning on build.

JorindeUsMedia (approved):

jpluscplusm (changes requested):
Please could you add a brief how-to-test to the commit message? I /think/ I know the places in the workflow where I'd look for now-missing error messages, but it'd be good to have it recorded.

JorindeUsMedia:
> Please could you add a brief how-to-test to the commit message? I /think/ I know the places in the workflow where I'd look for now-missing error messages, but it'd be good to have it recorded.

@jpluscplusm This is a main hugo version update so the main way to test is to see if the build is still working which it is.
Also previously we've got warnings showing in the hugo serve step: these are now gone.  
I don't think the build script in the trybot shows hugo warnings because debug is not turned on there. But when you run the project locally on current alpha you get hugo warnings when running serve: these warnings should be gone when running serve on this branch. 

Also the readme file was outdated and had v0.111.3 hugo version listed while the pipeline/build was using 115. Also we're using newer hugo stuff so the build/serve wasn't working anymore when running a version previous to 115. Since 117 is now the current hugo version we decided to update 117 while fixing the build warnings.

myitcv (approved):

myitcv:
Closed via f9c0894cbf18e3cf27585112a9db900218da3b8f

PR #397, opened 2023-08-22
hugo: Fix js error which triggers when codeToCopy is set to false

JorindeUsMedia:
- Only add data-copy when showCopyButton is set to true to avoid loading the copy javascript when it's not needed (and not working)

To test: Go to '/examples/basic/code-block/'. Previously this showed an error
in the console for codeblock with codeToCopy set to false. Now the error is
gone.

For https://linear.app/usmedia/issue/CUE-311

### DIFF
--- a/hugo/layouts/_default/_markup/render-codeblock.html
+++ b/hugo/layouts/_default/_markup/render-codeblock.html
@@ -4,7 +4,8 @@
 {{- $showCopyButton := ne (index .Attributes "codetocopy") false -}}
 {{- $codeToCopy := index .Attributes "codetocopy" | default (.Inner | base64Encode) -}}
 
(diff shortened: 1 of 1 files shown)


jpluscplusm (changes requested):
Please could you add a brief how-to-test to the commit message? I /think/ I can intuit my way to a passing test, but it'd be good to have it recorded somewhere apart from the PR.

jpluscplusm (approved):

jpluscplusm:
Imported as https://review.gerrithub.io/c/cue-lang/cuelang.org/+/1167809

jpluscplusm:
Closed via ec4ad92f6b9178512cbc610c1f28ed89013d3aba.

PR #398, opened 2023-08-22
Hugo: new header for article pages (docs, example, blog)

anne-usmedia:
- Added (headless) authors
- Author styling + popup
- Header styling for all article pages
- Breadcrumb styling on docs pages
- Fix things that broke because of new styling, like spacings, anchor links, separator lines
- Added authors example page with explaination (/examples/basic/authors/)

Notes:
- tags will be inline with the authors and meta data IF they all fit. When the tags do not all fit, they (in total) will move to the next line.
- authors have a fallback image when there is no image.
- the pointer in the author popup will only show from screen size 768px because of positioning issues. The mobile version of this might need some rethinking in general, again because of positioning difficulties.

Example pages
- /docs/howto/encode-json-yaml-with-cue/
- /blog/example/
- /examples/basic/front-matter/

For
- https://linear.app/usmedia/issue/CUE-260
- https://linear.app/usmedia/issue/CUE-275


content/docs/howto/encode-json-yaml-with-cue/en.md
```diff
@@ -3,9 +3,17 @@ title: Encode JSON or YAML with CUE
 weight:
 draft: false
 toc_hide: true
+date: "2023-08-22"
 tags:
     - Use encodings in CUE
+authors:
```
JorindeUsMedia, line 9: Not sure if we should already add this to existing content. Maybe only add it to the blog example and the front-matter example pages? Or we can add it to an existing docs page for testing but we need to check with Paul @myitcv  if we've added the right authors. Also I don't think reading time is relevant here (should be supported here but I don't think it will be just on docs that often.)

hugo/assets/scss/components/authors.scss
```diff
@@ -0,0 +1,182 @@
...
+    }
+
+    &--popover {
+        --author-img-size: 45px;
+
+        background-color: var(--author-popover-bg-color);
+        border: 1px solid var(--author-popover-border-color);
+        border-radius: $b-radius--large;
+        box-shadow: 0 10px 14px transparentize($c-blue--darker, 0.8);
+        cursor: auto;
```
JorindeUsMedia, line 67: is there a reason we have to set the cursor to a non-default value?

hugo/assets/scss/components/article.scss
```diff
@@ -55,8 +35,49 @@
...
+    &__footer {
+        display: flex;
+        flex-direction: column;
+        gap: 2rem;
+        margin: 0 auto;
+        opacity: 1;
+        padding: 1.5rem $p-gutter;
+
+        &::before {
+            border-top: 1px solid $c-lavender--light;
```
JorindeUsMedia, line 73: why not have the border on the __footer directly instead of pseudo elements?

hugo/assets/scss/components/authors.scss
```diff
@@ -0,0 +1,182 @@
...
+            background: linear-gradient(var(--author-text-color), var(--author-text-color)) no-repeat 100% 100%;
+            background-size: 0 1px;
+            display: inline;
+            transition: background-size 0.3s ease-in-out;
+        }
+    }
+
+    @include screen($screen-simple) {
+        &__item {
+            cursor: pointer;
```
JorindeUsMedia, line 151: cursor: pointer was allready set I think so no need to redefine it?

hugo/assets/scss/components/list.scss
```diff
@@ -47,6 +47,10 @@
                 grid-column: span 2;
             }
         }
+
+        // &--tags {
```
JorindeUsMedia, line 51: commented out can be removed I think? if we need it back we an use git history

hugo/content/en/authors/marcelvanlohuizen/mpvl.jpeg
JorindeUsMedia, line 1: just a note here: we need to think of a way to handle images and repo size. When we're going to have lots of images and have them both in the content & hugo content we might reach the github repo size limits.. Something to think of @myitcv  I will create a new linear for this

hugo/layouts/docs/single.html
```diff
@@ -8,11 +8,10 @@
     <div class="docs" data-docs>
         <div class="docs__main">
             {{ partial "paging/breadcrumb.html" . }}
-            {{ partial "separator.html" }}
 
             <article class="article article--docs">
                 <div class="article__container">
-                    {{ partial "article--header" . }}
+                    {{ partial "article--header" (dict "showTitle" false "context" .) }}
```
JorindeUsMedia, line 14: with the design with breadcrumb above the title: shouldn't the title always show as the article_title?

hugo/layouts/partials/authors.html
```diff
@@ -0,0 +1,67 @@
...
+
+<ul class="authors">
+    {{ range $index, $author := $authorsPages }}
+        {{ $cleanAuthorName := replace .Name "/index.md" "" }}
+        {{ if in $articleAuthors $cleanAuthorName }}
+        {{/*  sort the authors? Based on what?  */}}
+            <details class="authors__item" data-dropdown>
+                <summary class="authors__avatar">
+                    {{ if .Params.image }}
+                        <img class="authors__image" src="/{{ $author.File.Dir }}{{ .Params.image }}">
```
JorindeUsMedia, line 13: alt is missing

hugo/layouts/partials/authors.html
```diff
@@ -0,0 +1,67 @@
...
+                        <span class="authors__display-name">{{ .Params.displayName }}</span>
+                    {{ else }}
+                        <span class="authors__author-name">{{ .Params.name }}</span>
+                    {{ end }}
+                </summary>
+
+                <div class="authors--popover">
+                    <div class="authors__avatar">
+                        {{ if .Params.image }}
+                            <img class="authors__image" src="/{{ $author.File.Dir }}{{ .Params.image }}">
```
JorindeUsMedia, line 28: alt is missing

hugo/layouts/partials/authors.html
```diff
@@ -0,0 +1,67 @@
+{{ $authorsPages := .Site.GetPage "/authors" }}
+{{ $authorsPages := $authorsPages.Resources.ByType "page" }}
+{{ $articleAuthors := .Params.Authors | default false }}
+
+<ul class="authors">
```
JorindeUsMedia, line 5: this should not a ul because ul can only contain list items 
or the dtails element should be wrapped inside an li , or don't use ul here

hugo/layouts/partials/article--header.html
```diff
@@ -1,26 +1,41 @@
...
+{{ with $context }}
+    {{- if .Params.tags -}}
+        {{ $tagOptions := .Site.Params.Tags | default slice }}
+        {{ $tags = where $tagOptions "name" "in" .Params.tags | default slice }}
     {{- end -}}
-</header>
+
+    <header class="article__header">
+        {{ if $showTitle }}
+            <h1 class="article__title">{{ .Title }}</h1>
```
JorindeUsMedia, line 13: not sure if title should always be article__title instead of sometimes being part of breacrumb. Is there a reason it's part of the breadcrumb in docs? (seo maybe?)

hugo/assets/scss/components/article.scss
```diff
@@ -69,38 +90,45 @@
         max-width: $w-content;
         padding-top: 0;
 
+        #{ $self }__info {
+            justify-content: space-between;
```
JorindeUsMedia, line 94: in design the authors + meta is on the left and the tags on the right. 
Right now the meta is alligned in the center instead of the right on small screens for docs. 
Screenshot from 2023-08-22 16-46-31
On blogs it does look like design (on left)

JorindeUsMedia (changes requested):
Overall it already looks really nice! Just a couple of comments and/or questions.

content/docs/howto/encode-json-yaml-with-cue/en.md
anne-usmedia: For now it's to show how it all looks. If needed all can be removed before merging this PR @myitcv 
Of course, we can also add the correct values, but then I need to get them first ;)

hugo/layouts/partials/article--header.html
anne-usmedia: Do you prefer to use the breadcrumb for the title (like it is now) or the article title. In case of the latter, this needs some more work.

JorindeUsMedia: There's something to say for both ways. (seo, schema etc) For now just lets keep it as you have it now.

hugo/layouts/partials/authors.html
```diff
@@ -0,0 +1,67 @@
...
+                        {{ else }}
+                            {{- partial "icon.html" (dict "icon" "cue" "class" "authors__image") -}}
+                        {{ end }}
+
+                        <div class="authors__name">
+                            {{ if .Params.displayName }}
+                                <span class="authors__display-name">{{ .Params.displayName }}</span>
+                            {{ end }}
+
+                            {{ if .Params.name }}
```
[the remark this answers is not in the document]
anne-usmedia: No, this is in the popover, according to the design both need to be shown:
Screenshot 2023-08-23 at 10 32 10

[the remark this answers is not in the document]
JorindeUsMedia: ah sorry, my mistake (had an earlier design in mind..)

hugo/layouts/partials/authors.html
```diff
@@ -0,0 +1,67 @@
...
+                                <span class="authors__author-name">{{ .Params.name }}</span>
+                            {{ end }}
+                        </div>
+                    </div>
+
+                    <div class="authors__links">
+                        {{ if .Params.githubLink }}
+                            <div class="authors__github">
```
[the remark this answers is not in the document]
anne-usmedia: yes, but adding the icon inside the `a` will break the hover animation when the text falls over 2 lines. Or we need to think of another hover animation (there is no design for this)

hugo/assets/scss/components/article.scss
anne-usmedia: Because of paddings and margins etc.,  and the border would be too wide. See example below. The red one is when the border-top would be added directly to the `__footer`
Screenshot 2023-08-23 at 11 40 33

hugo/layouts/partials/authors.html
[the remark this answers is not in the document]
anne-usmedia: Moved some stuff around as discussed with Jorinde, so now the icons can be inside the `a` and thus are clickable

hugo/assets/scss/components/authors.scss
```diff
@@ -0,0 +1,190 @@
...
+    --author-text-color: #{ $c-blue };
+
+    align-items: center;
+    display: flex;
+    flex-wrap: wrap;
+    gap: 0.75rem $p-gutter;
+    position: relative;
+    width: fit-content;
+
+    &__item {
```
JorindeUsMedia, line 23: I'm missing a hover/focus state on the item: Maybe change the color of the name or hover?

JorindeUsMedia (changes requested):
Just 1 small thing (hover state on author item). Everything else looks great now!

hugo/assets/scss/components/authors.scss
```diff
@@ -0,0 +1,199 @@
...
+                #{ $self }__author-name {
+                    text-decoration: none;
+                }
+            }
+        }
+
+        &:focus,
+        &:focus-within,
+        &:hover {
+            background-color: var(--author-bg-color, $c-lavender--light);
```
JorindeUsMedia, line 45: looks better, just missing the 'transition' to make the animation run smoothly

JorindeUsMedia (changes requested):

JorindeUsMedia (approved):

content/docs/howto/encode-json-yaml-with-cue/en.md
myitcv: Thanks for flagging. We can leave for now, and we will tidy up later.

myitcv (approved):
This looks great. Importing into Gerrit now.

myitcv:
Closed via 2e6e47bc172476097ca18589f986d7834de74711

PR #399, opened 2023-08-23
Hugo: content footer redesign

anne-usmedia:
This changes the article footer on all single pages.

- New icons have been added for share and navigation (prev/next)
- All translations have been added.
- Copy functionality has been added, and copy.ts has been made more dynamic (also a few adjustments in the render-codeblock.html file as a result).
- The article--footer (partial) HTML has been added to all single pages.
- The share (partial) has been added.
- The tag--large modifier has been removed.
- The "last modified" is shortened to only show the "last modified" text + date. No more commit text.
- Arrow navigation is improved with new styling. Mobile now shows the arrows in stead of nothing.

Examples:
- /docs/howto/validate-yaml-using-cue/
- /blog/example/
- /examples/basic/front-matter/


hugo/assets/scss/components/button.scss
```diff
@@ -57,6 +57,7 @@
     }
 
     &--icon {
+        min-width: var(--button-height);
```
JorindeUsMedia, line 60: width wins from minwidth so I think this line is not needed

hugo/assets/scss/components/prevnext.scss
```diff
@@ -2,105 +2,87 @@
...
+    --prevnext-color: #{ $c-black };
+    --prevnext-icon-color: #{ $c-blue };
+    --prevnext-icon-size: 30px;
 
-    padding: 1.5rem 0;
+    padding-bottom: 2rem;
     position: relative;
 
     &__container {
         @include container($w-content);
```
JorindeUsMedia, line 19: we might need to change the container width to whatever docs has container width because now the prevnext arrow are not aligned with the rest of the footer. In design they are aligned with the footer  (yellow line)
Screenshot from 2023-08-25 10-16-57

hugo/assets/scss/components/share.scss
```diff
@@ -0,0 +1,124 @@
...
+
+        color: var(--text-link-color);
+        display: flex;
+        height: 100%;
+        padding: 0 0.5rem;
+        width: 100%;
+    }
+
+    &__copy,
+    &__link {
```
JorindeUsMedia, line 64: I'm missing hover/focus styles on the link and copy button

hugo/assets/svg/ui/social-weixin.svg
JorindeUsMedia, line 1: This can be removed I think since we can't share on wechat

hugo/assets/ts/widgets/share.ts
JorindeUsMedia, line 1: everything is commented out so this file can be removed?

hugo/layouts/partials/share.html
```diff
@@ -0,0 +1,35 @@
+<details class="share" data-dropdown>
+    <summary class="share__container">
```
JorindeUsMedia, line 2: the naming is a bit weird since this container only contains the button
also in the share.scss I see styles set to list-reset but this is not a list? so maybe both styling and clsasname can be removed and button classnames can be lifted to summary since the a itself doesnt do anything (see comment below)

hugo/layouts/partials/share.html
```diff
@@ -0,0 +1,35 @@
...
+        <a class="button button--icon button--light-blue share__button">
```
JorindeUsMedia, line 3: what is the reason we have an a inside a summary? The a doesnt link to antyhing: the summary is the thing thats clickable/focusable. Also a without a href is bad practice since it doesn't do anything. If you need antyhing for javascript purposes use a <button> but in this case I think all click events happen on the summary element so I think you can just remove this a and move all classnames to the summary element

hugo/layouts/shortcodes/url.html
```diff
@@ -0,0 +1 @@
+{{ .Permalink }}
```
JorindeUsMedia, line 1: what do we need this shortcode for? .Permalink is so simple you dont need to have a short code for it.. Also hugo has its own build in shotcodes for references and such. So I think this shortcode can be removed and just use {{ .Permalink }}in the template where it needed.

JorindeUsMedia (changes requested):

JorindeUsMedia (approved):

jpluscplusm (approved):
Importing now.

jpluscplusm:
Imported as https://review.gerrithub.io/c/cue-lang/cuelang.org/+/1168107

jpluscplusm:
Closed via 7715fe0a0cd3f42b027c049c9e4a1ca680e0b407

PR #401, opened 2023-08-25
hugo: LHS nav: show active item in tree

JorindeUsMedia:
Ignore the param 'toc_hide: true' for active items so the active item shows up in the nav-tree again because:
1. In order the see the table of contents the item needs to show in the left hand nav
2. To make clear where the active item lives in the current tree.

To test:
- Click on How-To guides: The page 'Encode JSON or YAML with CUE' doesn't show up here because toc_hide is set to true.
 - Now go to this page  (/docs/howto/encode-json-yaml-with-cue/): the page now does show up in the lhs nav because it's active and also the table of contents is showing again (in current alpha when going to the same page it does not show up in the nav and therefore the table of contents is also not visible)

For https://linear.app/usmedia/issue/CUE-282

### DIFF
--- a/hugo/layouts/partials/paging/tree.html
+++ b/hugo/layouts/partials/paging/tree.html
@@ -53,7 +53,13 @@
     {{ $isActive := (eq $s $p) -}}
     {{ $isActivePath := ($p.IsDescendant $s) -}}
 
-    {{ $pages := where (union $s.Pages $s.Sections).ByWeight ".Params.toc_hide" "!=" true -}}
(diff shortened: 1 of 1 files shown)


anne-usmedia (approved):

myitcv (approved):
Perfectly timed, thanks. I was going to do some hacking this evening to get the LHS nav into the shape we want.

Importing into Gerrit now.

myitcv:
Closed via ac3925d5c0ea0560679f84aaf19046995ad919bc

PR #402, opened 2023-08-25
playground: Policy workspace

JorindeUsMedia:
- Enable policy workspace + add config
- On load: determine workspace first: then load params from url + do validation
on params & config
- Move code for url params, setting up workspace, wasm and share to separate
helpers to avoid having a massive file and to split logic to keep everything
understandable and easy to edit
- Support both old (split by '@') and new way (named params:
w:function/i:cue/f:def/o:yaml) of getting params from url
- Move getting saved code to before load of editor instead of after (faster)
- Show loader when workspace is still loading
- Update hash when hash is different because of validation rules but use replace
 when coming from load/update to avoid triggering another hash change which
 would trigger update again.
- Adjust styling so it works with both function (2 columns, 1 row) and policy
workspace (2 columns, 3 rows per column)
- Dynamic render the workspace config (inputs tabs, func & output)
- Change workspace-menu config to update the playground on change instead of on
submit because this makes more sense ux wise. Now you can immediately see what
changes a setting does.
- Prepare wasm types & wasm helper & output function on the app to work with
multiple-input workspaces like policymode. Comment out for now until wasm part
is implemented.
- Prepare output code to set the about language depending on the wasm result
- Add readonly option to output and dropdowns so user cant change it: can only
be changed via output of wasm.
- Prepare share code for multi-inputs sharing (for now it just handles function
 workspace but can easily be changed to handle tartxt result)
- Disable share for now for policy mode
- Update gitignore to ignore .idea folders everywhere, not just root
- Make sure key's of list items are unique for that item in the list
- Fix imports so they al use the custom paths from tsconfig.json

To test: Go to /play and click around with the workspace options. Switch the
workspace via the workspace menu at the top.
Note: WASM for Policy workspace is not working correctly yet.
Note: Share is temporarily disabled for policy workspace now.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,7 +3,7 @@ node_modules
 /.nvmrc
 
 # IDE's
-/.idea
+.idea
 .vscode/
 
 # Project
--- a/playground/src/components/dropdown.tsx
+++ b/playground/src/components/dropdown.tsx
@@ -1,15 +1,16 @@
 import * as React from 'react';
 import { MouseEvent, MutableRefObject } from 'react';
 import cx from 'classnames';
-import { Icon } from './icon';
-import { DropdownChange, DropdownItem } from '../models/dropdown';
+import { DropdownChange, DropdownItem } from '@models/dropdown';
+import { Icon } from '@components/icon';
 
 interface DropdownProps {
     activeItem: DropdownItem;
     cssClass?: string;
     disabled?: boolean;
+    readonly?: boolean;
     disabledText?: string;
-    id: string;
+    groupId: string;
     items: DropdownItem[];
     name?: string;
     onDropdownSelect?: {(change: DropdownChange): void};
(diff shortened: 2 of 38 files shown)


myitcv (approved):
This looks really good, thanks.

> Change workspace-menu config to update the playground on change instead of on
submit because this makes more sense ux wise. Now you can immediately see what
changes a setting does.

One thing I think we might end up doing is not requiring a press of "OK" when something is changed. For some reason, I find myself expecting that a click to change the workspace will dismiss the dialog and switch the workspace. It's always a surprise to me that I need to click ok.

Will import into Gerrit now though and we can revisit this.

myitcv:
Closed via f8f99a2c8e63901a583a19c94adf1971bd091153

JorindeUsMedia:
> This looks really good, thanks.
> 
> > Change workspace-menu config to update the playground on change instead of on
> > submit because this makes more sense ux wise. Now you can immediately see what
> > changes a setting does.
> 
> One thing I think we might end up doing is not requiring a press of "OK" when something is changed. For some reason, I find myself expecting that a click to change the workspace will dismiss the dialog and switch the workspace. It's always a surprise to me that I need to click ok.
> 
> Will import into Gerrit now though and we can revisit this.

Yes agreed. The 'OK' now only closes the menu. I would still recommend some button to close it but maybe rename it to 'Close' or 'Done'

PR #409, opened 2023-08-29
Hugo: new search results styling

anne-usmedia:
- Added breadcrumb to algolia.json
- Added the breadcrumb to the search result teasers
- Adjusted the styling according to design
- Adjusted Algolia settings: 
    - use `<strong>` in stead of `<mark>` for highlighting 
    - cut summary after 30 words in stead of 10

For testing:
1. create a new test index in Algolia (see hugo > readme for a walkthrough)
2. (temporarily) replace the index names with your new index name
3. look at the search results, both the search widget as on the /search page

For testing the styling changes:
- For example, search for `JSON` or `blog` on the search page
- For example, search for `JSON` or `blog` in the search widget

Screenshot 2023-08-29 at 17 16 10

For https://linear.app/usmedia/issue/CUE-290

### DIFF
--- a/hugo/algolia-settings.js
+++ b/hugo/algolia-settings.js
@@ -24,9 +24,9 @@ index.setSettings({
     attributesForFaceting: ['tags', 'contentType'],
     distinct: true,
     attributeForDistinct: 'link',
-    highlightPreTag: '<mark>',
-    highlightPostTag: '</mark>',
-    attributesToSnippet: ['summary'],
+    highlightPreTag: '<strong>',
+    highlightPostTag: '</strong>',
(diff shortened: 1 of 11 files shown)


JorindeUsMedia (approved):
Look good :) 
For testing: I know algolia index & settings changes will only be updated when merged to alpha.. Maybe add a screenshot where we can see the result of what the breadcrumbs look like?
The tags restyling and other style changes of the search teaser **can** be tested via the preview link so maybe also add that to the commit 'to test' part.

myitcv (approved):
LGTM - will import into GerritHub now.

PR #420, opened 2023-09-01
playground: Examples

JorindeUsMedia:
- Make sure config objects are immutable by using DeepFreeze helper because Objec.freeze only is shallow but we need deep because of nested config properties.
- Use lodash deepclone to deep clone workspaces & example because we don't want to edit the original object which did happen before because in js object are reference values. Before this was not an issue because the config was only used for workspaces in the playground, but now it's used for example too so this was a problem now with multiple places editing the original config which should not
 happen.
- Rename hash-params helper to url-params, so we can have helper functions for not just hash params but also search params there
- Move updating hash & updating search params to the url param helper
- On load of page get example params from url and load example (note: if we have both id (saved code) and example in the url: id wins)
- Update searchparams via helper when needed
 - Add example menu component
 - Use Fuzzy search in example menu to filter the array of examples
 - Group the examples by category and sort by category weight
 - Add styling for search menu, input & items

To test: Go to /play. Click on examples. Try the search input and click on an example to test it. The url should update and the example should be loaded

For https://linear.app/usmedia/issue/CUE-227

### DIFF
--- a/playground/package-lock.json
+++ b/playground/package-lock.json
@@ -9,12 +9,15 @@
       "version": "0.1.0",
       "dependencies": {
         "@monaco-editor/react": "4.5.1",
+        "fuse.js": "6.6.2",
+        "lodash.clonedeep": "4.5.0",
         "lodash.debounce": "4.0.8",
         "react": "18.2.0",
         "react-dom": "18.2.0",
         "react-scripts": "5.0.1"
       },
       "devDependencies": {
+        "@types/lodash.clonedeep": "4.5.7",
         "@types/lodash.debounce": "4.0.7",
         "@types/react": "18.0.28",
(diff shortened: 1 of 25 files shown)


jpluscplusm (approved):
> To test: Go to /play. Click on examples. Try the search input and click on an example to test it. The url should update and the example should be loaded

This works for me, but I'm *utterly* unable to review anything under the hood.

Leaving this until it gets a TS-/JS-capable person's review!

myitcv (approved):
This looks great. Will import into GerritHub now.

We can tweak the actual examples later.

myitcv:
Closed via 78826eff1675fede929e31629ad76fdf22b4a905

PR #422, opened 2023-09-01
hugo: add search shortcode

JorindeUsMedia:
- Add search shortcode
- Move search to a partial to be able to reuse it
- Add styling for search--widget
- Add logic for hidden field is contentTypes dropdown is hidden
- If content type is filled in: use this as placeholder in the searchbar
- Remove check on empty query because we also want to search when no query is
filled in but facets are set
- Fix scoping issue for search ts (checking within element instead of
document)
- Change handle submit to use current page when type is results
- Add example page
- Fix issue with blog having been renamed to 'Blog overview' and because it's
a list page this was also renamed in the search results and filters

To test: examples/shortcodes/search

For https://linear.app/usmedia/issue/CUE-283

### DIFF
--- a/content/blog/_en.html
+++ b/content/blog/_en.html
@@ -1,5 +1,5 @@
 ---
-title: "Blog overview"
-description: "Blog overview description."
+title: "Blog"
+description: "CUE blogs"
 layout: list
 ---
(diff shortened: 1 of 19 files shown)


myitcv (approved):
LGTM - will import into GerritHub now.

myitcv:
Closed via e71ca6a79d62c044aaf8767ac0ac7770816026a3

PR #424, opened 2023-09-07
Update the brew install directions

Senjai:
Cue has a formula in homebrew core, which is probably what should be used unless I'm missing something.

https://formulae.brew.sh/formula/cue


myitcv:
Thanks for raising this, and apologies for dropping the ball on this and other PRs in this repo.

I think we will keep with the tap for now, because that is more clearly under our control. But we should pick up the conversation in https://github.com/cue-lang/cue/issues if that is not the correct position.

PR #426, opened 2023-09-18
content/hugo: Fix problem with authors not showing

JorindeUsMedia:
- Rename authors/index.html to en.md because html is not needed here since it's headless. Also we need 'en' instead of 'index' here in order to make this work with multiple languages & the preprocessor.
- Add a cue file to the index file for authors, so it also gets copied to hugo/content folder.
- Add .Params.authors to the if statement in article-header so the info also gets rendered when we just have authors and no meta and no date
- Change the range within the authors partials to loop through .Params.authors instead of all authors: this way we keep the order of the authors as set in the front-matter. The other way around it always gets sorted by foldername which we don't want
 - Get full author page from the author-param-string by using Resources.Match which is the way to get pages on Headless Content types since the regular getPage way doesn't work here.
 - Remove inconsistencies with using $author. and . by just using . (dot) everywhere in the authors partial.
 - Remove marcel from the example blog authors, so we can better test getting the authors from the front-matter. Previously they were the same as all authors so a bit hard to test if the range and sorting logic really worked.

To test: /blog/example

For https://linear.app/usmedia/issue/CUE-332


jpluscplusm (approved):
Example page looks great 👍

jpluscplusm:
Imported as https://review.gerrithub.io/c/cue-lang/cuelang.org/+/1169531

jpluscplusm:
Closed via 7cb1abfd93ec6822c46e822647c06b745118a094

PR #427, opened 2023-09-20
hugo: Fix search results bug when tag is not found in tag config

JorindeUsMedia:
- If a tag on a page is not found in the tag config: show it anyway using the tag-string as name and default color 'blue'

To test: Go to /search and search for 'yaml'. Search results show again and the tag 'yaml' is displayed with the default color blue.

For https://linear.app/usmedia/issue/CUE-336

### DIFF
--- a/hugo/assets/ts/widgets/search-results.ts
+++ b/hugo/assets/ts/widgets/search-results.ts
@@ -98,7 +98,10 @@ export class SearchResults extends BaseWidget {
     public createTeaser(teaser: Teaser): string {
         let tags = null;
         if (teaser.tags && teaser.tags.length) {
-            tags = teaser.tags.map(tagString => this.tags.find(tag => tag.name === tagString));
+            tags = teaser.tags.map(tagString => {
+                const fullTag = this.tags.find(tag => tag.name === tagString);
+                return fullTag ?? { name: tagString, color: 'blue' };
+            });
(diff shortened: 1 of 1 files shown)


jpluscplusm (approved):
This fix achieves what it says, but in doing so we now have 2 different behaviours:

- a search will show a tag (that's missing from the tag config); but
- a click through to such a page displayed in search **won't** show the "missing" tag.

Approving and importing; but tagging @myitcv for a decision on if we need to fix the second of those behaviours as well, for consistency.

jpluscplusm:
Imported as https://review.gerrithub.io/c/cue-lang/cuelang.org/+/1169706

JorindeUsMedia:
> This fix achieves what it says, but in doing so we now have 2 different behaviours:
> 
> * a search will show a tag (that's missing from the tag config); but
> * a click through to such a page displayed in search **won't** show the "missing" tag.
> 
> Approving and importing; but tagging @myitcv for a decision on if we need to fix the second of those behaviours as well, for consistency, in a future PR.

Every used tags should be configured in the params.toml to avoid this behaviour. But for now this fix makes sure the search results won't break when it's not in there

jpluscplusm:
Closed via 81a0584fe8de7f62cfa1b778a2ca0f20f9cec455

PR #434, opened 2023-11-14
Hugo: moved share in article footer to top

anne-usmedia:
Moved the share button in article footer to top when there are no tags. Now it is at the same height as the "last modified".

For testing, scroll down to footer in eg. /docs/howto/about-these-guides/

For https://linear.app/usmedia/issue/CUE-342

### DIFF
--- a/hugo/assets/scss/components/article.scss
+++ b/hugo/assets/scss/components/article.scss
@@ -99,6 +99,11 @@
     &__share {
         flex: 0 0 46px;
         height: 46px;
+
+        &:only-child {
+            margin-left: auto;
+            margin-top: -3.75rem
+        }
     }
 
     &--blog {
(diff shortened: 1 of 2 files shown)


hugo/assets/scss/components/article.scss
```diff
@@ -99,6 +99,11 @@
     &__share {
         flex: 0 0 46px;
         height: 46px;
+
+        &:only-child {
+            margin-left: auto;
+            margin-top: -60px;
```
JorindeUsMedia, line 105: maybe do this in rem since it's (I guess) based on the font-size of 'last-modified'. If a user for instance has a bigger font in the browser (accessibility) using rem's it will have the margin top bases on the font size

anne-usmedia: It's not, but I'll change it to rem anyway

JorindeUsMedia (approved):

myitcv:
Closed via 691c431a2aa470094062607493133ba2d4ec4c04

PR #439, opened 2023-11-15
Playground: do not index or crawl playground links with 'id'

anne-usmedia:
- Added Disallow: /play/*?id= to robots.txt

For https://linear.app/usmedia/issue/CUE-289

### DIFF
--- a/hugo/layouts/robots.txt
+++ b/hugo/layouts/robots.txt
@@ -1,4 +1,5 @@
 User-agent: *
 Allow: /
+Disallow: /play/*?*id=
 
 Sitemap: {{ absURL "sitemap.xml" }}


playground/src/containers/app.tsx
```diff
@@ -431,9 +431,12 @@ export class App extends React.Component<AppProps, AppState>
         const activeWorkspace = this.state.workspaces[this.state.activeWorkspaceName];
         const contents = workspaceToShareContent(activeWorkspace, this.inputEditors);
         const data = await share(contents);
+
         if (data) {
             window.history.pushState({}, 'CUE Playground', '?id=' + data + window.location.hash);
             this.setState({ ...this.state, saved: true, showSaveURL: true });
+
+            updateRobotsMetaNoIndex(new URLSearchParams(window.location.search), 'id');
```
JorindeUsMedia, line 439: This is now adding the meta tag after it's been shared to the current page. When a person clicks on a shared link nothing will be added because they don't share it themselves: they just open the link. so if we want to do something like this we need to do it on page load. Also I'm not sure if this will work at all since it probably executes after the robots have already done their thing but it's worth a try.. But we do need something on load: not on share.

JorindeUsMedia (changes requested):
I don't know if this will work since our javascript probably executes after the crawler has already crawled the page. The part in robots.txt should be working but the parts in the app.tsx and url-params probably not because they happen when the bot has already  done it's job.. If we want do something like this we have to do it when the page loads and not when its shared.

JorindeUsMedia (approved):

ErikHendrikson (approved):

anne-usmedia:
> I don't know if this will work since our javascript probably executes after the crawler has already crawled the page. The part in robots.txt should be working but the parts in the app.tsx and url-params probably not because they happen when the bot has already done it's job.. If we want do something like this we have to do it when the page loads and not when its shared.

The JS part has been removed

playground/src/containers/app.tsx
```diff
@@ -431,6 +431,7 @@ export class App extends React.Component<AppProps, AppState>
         const activeWorkspace = this.state.workspaces[this.state.activeWorkspaceName];
         const contents = workspaceToShareContent(activeWorkspace, this.inputEditors);
         const data = await share(contents);
+
```
jpluscplusm, line 434: This looks unrelated!

playground/src/helpers/url-params.ts
```diff
@@ -34,6 +34,7 @@ export const getHashParamsFromUrl = (): hashParams => {
 
 export const updateHash = (newHash: string, replace = false): void => {
     const currentHash = window.location.hash.slice(1);
+
```
jpluscplusm, line 37: This looks unrelated!

hugo/layouts/robots.txt
```diff
@@ -1,4 +1,6 @@
 User-agent: *
 Allow: /
 
+Disallow: /play/*?id=
```
jpluscplusm, line 4: This should be `/play/*?*id=` in order to catch `id` no matter where it appears in the query string order, which will be useful as more params are added (or if we already have additional params in URLs).

jpluscplusm (changes requested):

hugo/layouts/robots.txt
```diff
@@ -1,4 +1,6 @@
 User-agent: *
 Allow: /
```
jpluscplusm, line 3: The well-hidden but mostly-canonical reference site for `robots.txt` mentions blank lines being a problem:

> you may not have blank lines in a record, as they are used to delimit multiple records.

-- https://www.robotstxt.org/robotstxt.html

jpluscplusm (changes requested):

playground/src/containers/app.tsx
anne-usmedia: Ah yes, there used to be some JS which was later removed again. Will remove the extra whitespace 👍

hugo/layouts/robots.txt
anne-usmedia: Done

hugo/layouts/robots.txt
anne-usmedia: Fixed!

jpluscplusm (approved):
LGTM - importing now.

jpluscplusm:
Imported as https://review.gerrithub.io/c/cue-lang/cuelang.org/+/1172313

jpluscplusm:
Closed via 8dfb142ecf0635de2620656b4aff564e7d1642d5

PR #443, opened 2023-11-15
Typos: extra go code parens and extra 's'

ehershey:


myitcv:
@ehershey apologies for dropping the ball here too. I've picked up these changes in https://review.gerrithub.io/c/cue-lang/cuelang.org/+/1194128. In any case, as part of the upgrade to https://cuelang.org we also will update the content on this page. cc @jpluscplusm to confirm in which docs and content issue that will happen please?

PR #446, opened 2023-11-16
Hugo: remove deprecated social.twitter from config

anne-usmedia:
Removed the social key in the site configuration. Wanted to add it to the params.toml, but noticed the DO NOT EDIT note.

It looks like social.twitter is not used anywhere. If it is, it needs to be added to the params.toml and can be called with Site.Params.social.twitter

For https://linear.app/usmedia/issue/CUE-345

### DIFF
--- a/hugo/config/_default/config.toml
+++ b/hugo/config/_default/config.toml
@@ -51,9 +51,6 @@ disableAliases = false
     [mediaTypes."text/netlify"]
         delimiter = ""
 
-[social]
-    twitter = "cue_lang"
-
 [build]
     noJSConfigInAssets = true
 


JorindeUsMedia:
> Removed the social key in the site configuration. Wanted to add it to the params.toml, but noticed the DO NOT EDIT note.

This is not needed because the one in params.toml is about the twitter card and I think that one is still used for the twitter meta tags (when you share a page on twitter)

JorindeUsMedia (approved):

JorindeUsMedia:
> > > Removed the social key in the site configuration. Wanted to add it to the params.toml, but noticed the DO NOT EDIT note.
> > 
> > 
> > This is not needed because the one in params.toml is about the twitter card and I think that one is still used for the twitter meta tags (when you share a page on twitter)
> 
> Can you elaborate, since I do not understand what you want exactly.

I thought you weree talking about removing the twitter items in the params.toml : that is not needed. But I might have misread your comment because I think in the end it was ok to just delete the unused param

jpluscplusm:
@JorindeUsMedia @anne-usmedia To confirm my understanding of your conversation, before I import: this PR's removal of a Twitter-related key from `hugo/config/_default/config.toml` is correct and sufficient; we don't (currently) need to remove a similar key from any other files.

Have I understood correctly?

anne-usmedia:
@jpluscplusm Tthere was only a question if the twitter config had to be relocated from the config.toml to the params.toml. But since it is not being used, removing it from the config.toml is sufficient

jpluscplusm (approved):

jpluscplusm:
Imported as https://review.gerrithub.io/c/cue-lang/cuelang.org/+/1172308

jpluscplusm:
Closed via 4c42f40b91322868759dcc4c052183e3e88c3b0a

PR #449, opened 2023-11-17
Hugo: use dartsass instead of libsass

anne-usmedia:
- Installed (dart)sass
- Install Dart Sass on build on netlify https://gohugo.io/hugo-pipes/transpile-sass-to-css/#netlify
- Removed deprecated stylelint rule
- Fix linters for dartsass compatibility, mainly transparentize to color.scale

For: https://linear.app/usmedia/issue/CUE-296


JorindeUsMedia (changes requested):
I think we also need to tell hugo to use dart sass instead of their default: libsass
https://gohugo.io/hugo-pipes/transpile-sass-to-css/
I think that's the reason the pipeline is now failing because it's trying to compile newer sass functions using libsass now

myitcv:
FYI for anyone following this PR, I've started tracking this in https://github.com/cue-lang/cue/issues/2880

myitcv:
This is probably a good change to make, but I'll close this for now until we prioritise via Anne-Jet and Jorinde.

PR #450, opened 2023-11-17
Hugo: First stept for image processing

anne-usmedia:
Added processing for carousel and youtube images

Testing:
Both the carousel and youtube videos have a smaller size (800 x 450px) and are in the webp format (check via inspector mode)

For https://linear.app/usmedia/issue/CUE-103


JorindeUsMedia (approved):
Good first step for now to improve loading.  Can you also resize just by width so we can use that for content images which have different aspect ratios? 
Also would be nice if can create a srcset. 
Can you made follow up issues for the next steps? (teaser, content images, srcset etc)

myitcv:
This is probably a good change to make, but I'll close this for now until we prioritise via Anne-Jet and Jorinde.

PR #461, opened 2024-06-18
hugo: Fix search not working on mobile

JorindeUsMedia:
- Move the max width of the detached mode query to a variable so it can be
reused
- Move the on doc click logic to its own function
- Save if autocomplete is open in a variable so we can access it on doc click
- On click of the document only close & reset the algolia-autocomplete when
is not open AND click is outside element AND outside overlay (in detached mode
 overlay is outside the element)
- Fix issue with right side of search-icon in bar not showing on mobile
- Fix type issue with autocomplete: type was unknown so I've added the right
types

Fixes: https://github.com/cue-lang/cue/issues/2848

To test:
On mobile click on the menu (hamburger) and click in the search box. Type in a
 search query and either hit the enter key or click the search icon next to
 the input field.

### DIFF
--- a/hugo/assets/scss/components/searchbar.scss
+++ b/hugo/assets/scss/components/searchbar.scss
@@ -17,7 +17,7 @@
         border: none;
         border-radius: 50px;
         height: $h-button--small + 16px;
-        padding: 0 1.125rem 0 2.5rem;
+        padding: 0 1.125rem 0 2.75rem;
         position: relative;
 
         &:focus-within {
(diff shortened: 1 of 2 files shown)


JorindeUsMedia (commented):
comment to self:
there is also a not-detached mode version on the results page which now never closes 
should find a way to only target the detached one and not the results one

anne-usmedia (approved):

myitcv:
Thanks - reimported to https://review.gerrithub.io/c/cue-lang/cuelang.org/+/1196529

myitcv (approved):
LGTM

PR #463, opened 2024-06-21
playground: selected text isn't highlighted

JorindeUsMedia:
- Overwrite part of the vscode theme so the line highlight is transparent so the word highlight is also visible
- Turn off highlight active line by overruling part of the basic setup optionsof the code-mirror editor
- Remove css overrides & use vscodeLight terminal theme which overrides part of the vscode light theme
- Change selectionMatch color so it matches the selection color (blue instead of brown)
- Uncomment updateOutput function (should be on)

Fixes: https://github.com/cue-lang/cue/issues/3239

To test:
Go to /play
- Select a part of the code: it should be blue now
- Double-click a word: it should be blue instead of brown
- Do the same for the output editor
- Click on a line: line should not be highlighted

### DIFF
--- a/playground/src/config/editor-theme.ts
+++ b/playground/src/config/editor-theme.ts
@@ -1,0 +1,17 @@
+import { vscodeLightInit } from '@uiw/codemirror-theme-vscode';
+
+export const vscodeLight = vscodeLightInit({
+    settings: {
+        lineHighlight: '#f0f0f0b3',
+        selectionMatch: '#add6ff',
+    },
+});
+
+export const vscodeLightTerminal = vscodeLightInit({
+    settings: {
+        background: '#eef0f5',
+        gutterBackground: '#eef0f5',
+        lineHighlight: '#f0f0f0b3',
+        selectionMatch: '#add6ff',
+    },
+});
--- a/playground/src/containers/app.tsx
+++ b/playground/src/containers/app.tsx
@@ -13,8 +13,7 @@
 // limitations under the License.
 
 import { LanguageName, loadLanguage } from '@uiw/codemirror-extensions-langs';
-import CodeMirror, { EditorView, Extension } from '@uiw/react-codemirror';
-import { vscodeLight } from '@uiw/codemirror-theme-vscode';
+import CodeMirror, { basicSetup, EditorView, Extension } from '@uiw/react-codemirror';
 import { debounce } from 'lodash';
 import * as React from 'react';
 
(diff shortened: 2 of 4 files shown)


myitcv (approved):
LGTM

jpluscplusm (changes requested):
My initial impression is that whilst this improves the UX, there are now two different colours for selected text on the "current" line versus selected text on any /other/ line. I think the current-line colour should be aligned with the other-lines colour.

To see what I mean, open the preview at https://cl-1196613-1--cue-cls.netlify.app/play/#w=function&i=cue&f=eval&o=cue and swipe-and-select text from more than one line. Or just ctrl-A to select the whole pane - the last line will be highlighted in the current-line colour.

myitcv:
Nicely spotted, @jpluscplusm - this is indeed a problem. Appears to be a result of the current line highlight colour getting in the way (or mixing in some way with the selected text highlight). Perhaps we just turn off the current line highlight? It's a bit annoying, and with these sorts of weird interactions it's definitely not critical.

JorindeUsMedia:
I was also in doubt about just turning the highlight line off so I will do that. @myitcv @jpluscplusm

myitcv:
Now that my eye has been brought to the attention of issues in this space, one more tiny problem. Consider:

Screenshot 2024-06-24 at 11 35 01

There are two areas highlighting. Which one is selected, and which is highlighted as a result of the first?

Perhaps this isn't an issue in Visual Studio, but this seems confusing to me because if I were to hit delete it wouldn't be immediately obviously which would one or both would be deleted. 

@JorindeUsMedia - any experience from other editors here on what's normal?

JorindeUsMedia:
@myitcv We can make them a different color. From @jpluscplusm 's comment I understood you wanted them to be the same color.  But maybe I misunderstood because we had 2 problems: multi select vs select and line select (now turned off) vs manual select

myitcv:
Ok let's go with this for now and we can revisit after some local testing!

myitcv:
(I'll also have a look at what is causing the problems on your side with the PRs - the steps I outlined should have fixed the problem)

jpluscplusm:
I've re-imported the PR as of the most recent push and mailed it to https://cuelang.org/cl/1196613, replacing the previous CL so we can see a preview.

jpluscplusm (approved):
This looks good to me (preview at https://cl-1196613-2--cue-cls.netlify.app/play/#w=function&i=cue&f=eval&o=cue).

PR #465, opened 2024-07-17
playground: Playground gets stuck at "... loading WASM"

JorindeUsMedia:
- Fix check on inputEditors empty: it should check if not empty object instead
of undefined because default state is empty object.
- Change === undefined checks to use ! so it also covers null
- Move getting activeworkspace after the if statement because there is no need
to getting it when editors and wasm aren't ready
- Use react state to update value of output editor instead of transaction
because with transaction the change doesn't always update the state
correctly because it can conflict with other state changes of the editor.

Fixes: https://github.com/cue-lang/cue/issues/3271

To test:
Go to https://deploy-preview-465--cue.netlify.app/play/?id=6sgdmi8vArV#w=function&i=cue&f=eval&o=cueAfter loading the output editor should display the output code instead of
"loading wasm"

### DIFF
--- a/playground/src/containers/app.tsx
+++ b/playground/src/containers/app.tsx
@@ -46,6 +46,7 @@ interface AppState
     saved: boolean;
     showSaveURL: boolean;
     activeExample?: Example;
+    outputEditorValue?: string;
 }
 
 // App is the root of our React application
@@ -64,6 +65,7 @@ export class App extends React.Component<AppProps, AppState>
             saved: false,
             showSaveURL: false,
             activeExample: null,
+            outputEditorValue: undefined,
         };
     }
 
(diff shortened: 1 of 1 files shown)


myitcv (approved):
LGTM, thanks!

jpluscplusm:
@JorindeUsMedia Just for clarity, did the symptoms outlined in https://github.com/cue-lang/cue/issues/3271 directly lead you to the changes in this PR, or is this PR more speculative? I'm wondering how definitive you feel the fix is, given the issue.

(I'm **not** asking this because I've seen any repeat of the issue, but only because I'd like to know how confident we are that the underlying issue is understood, addressed, and shouldn't reoccur.)

JorindeUsMedia:
Yes it was based on the issue. After that I did a lot of debugging to find the real cause of the issue which were 2 things:

- The if statement to check if both editors & wasm were ready was sometimes skipped because it checked on undefined inputeditors while the default was {} (empty object) which caused that the playground tried to load the output value into an editor which was not ready
- After fixing that the issue still occurred so I did more debugging (added console log's to all the steps that could trigger an update of the output or an update of the editor) and found out that the dispatch of the transaction (update output value) got triggered correctly but after that 2 update events from the editor fired instead of 1 which stopped the first (our update). I tried to debug why it was doing that but found out by reading older issues in both react-codemirror and codemirror repository  it could be a lot of things. So instead of trying to "fight" this I decided to update the code via the app state instead of the editor state.

jpluscplusm:
@JorindeUsMedia Brilliant - thank you!

PR #466, opened 2024-07-17
hugo: Toc: always show parent pages of active page in tree

JorindeUsMedia:
When getting the page of a section also show pages which are ancestors of the current page. (page is descendant of page in loop). So effectively  ignore toc_hide for those pages

Fixes: https://github.com/cue-lang/cue/issues/3277

To test:
- Set toc_hide: to true on content/docs/reference/command/_en.md
- Go to /docs/reference/command/cue-help/ > "The cue command" item + tree should show
- Go to another page in the tree, for example /docs/reference/glossary > "The cue command" item should not show

### DIFF
--- a/hugo/layouts/partials/paging/tree.html
+++ b/hugo/layouts/partials/paging/tree.html
@@ -56,7 +56,8 @@
     {{ $pages := slice }}
     {{/* Only add pages that are active or have toc_hide set to false */}}
     {{ range (union $s.Pages $s.Sections).ByWeight -}}
-        {{ if or (eq . $p) (ne .Params.toc_hide true) -}}
+        {{ $pageIsActivePath := ($p.IsDescendant .) -}}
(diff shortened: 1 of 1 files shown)


myitcv (approved):
Great, thank you!

jpluscplusm (commented):
@JorindeUsMedia This is brill - thank you!

For posterity, and because we don't use this pattern anywhere yet (because it didn't display, until this PR!), I've rebased the WIP CL that demos the (now fixed) problem.

Here's a pair of example pages:
- a 4th-level page, inside a 3rd-level contain that has `toc_hide:true`: https://cl-1197613-2--cue-cls.netlify.app/docs/reference/command/cue-help/ -- the 3rd-level container is visible in the LHNav/TOC
- a 3rd-level page inside the same 2nd-level hierarchy as the previous example: https://cl-1197613-2--cue-cls.netlify.app/docs/reference/glossary/ -- the 3rd-level container sibling from the previous example is not visible in the LHNav.

Thanks again :-D

PR #467, opened 2024-07-17
hugo: Filter on hidden facet inputs for search widget

JorindeUsMedia:
- Move function go get facets from hidden inputs to helper so it can be used for both results & autocomplete
- Change function arguments of suggestions & autocomplete config to arrow functions so 'this' can be used. Move inline functions to the class itself.
- For searchbar of type results also get facets from the hidden inputs and not just from the url.
- Get search url for suggestions from window location when is widget; otherwise use main search url

 Fixes https://github.com/cue-lang/cue/issues/2841

 To test:
 - Go to https://deploy-preview-467--cue.netlify.app/docs/concept/find-a-guide
 - Type a search term into the bar.
 - When you click one of the suggestions you stay on the same page and search for the suggestion instead of going to the main search page
 - The suggested page items are within the concept guide section
 - When entering a term and hitting enter or the search button you stay on the same page and search for the entered term

### DIFF
--- a/hugo/assets/ts/helpers/search.ts
+++ b/hugo/assets/ts/helpers/search.ts
@@ -106,3 +106,23 @@ export const parseQuery = (query: string): ParsedQuery => {
         facets: cleanObject<SearchFacets>(facets),
     };
 };
+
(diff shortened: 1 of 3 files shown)


myitcv (approved):
LGTM, thank you!

PR #469, opened 2024-08-27
hugo: added select-and-copy filenames in code tabs

JuanUsMedia:
- Added small file icon (select-and-copy) prefix for filenames in code-tabs
- Added select-and-copy button for copying filename of code-tab
- Added file icon to ui.svg

Fixes: https://github.com/cue-lang/cue/issues/2935

To test:
- For all code-tabs with type "default" the file copy icon is added
- Default is shoing a file icon, this can be disabled by adding fileIcon="false" to shortcode code-tab For example: {{< code-tab name="problem1.cue" language="cue" area="top-left" fileIcon="false" >}}


content/examples/shortcodes/code-tabs/en.md
```diff
@@ -68,7 +68,7 @@ languages:
   name: Norwegian
 {{< /code-tab >}}
 
-{{< code-tab name="data.yaml" language="yaml" linenos="true" area="bottom-left" >}}
+{{< code-tab name="data.yaml" language="yaml" linenos="true" area="bottom-left" fileIcon="false" >}}
```
JorindeUsMedia, line 71: also update the code-isage example above with the same so the example and result are the same

hugo/layouts/partials/tabs.html
```diff
@@ -20,7 +20,14 @@
                 data-tab-item="{{ .name | urlize }}"
                 data-tab-group="{{ $groupId }}"
                 class="tabs-nav__tab{{ if ne .type "default" }} tabs-nav__tab--{{ .type}}{{ end}}{{ cond (eq $idx 0) " is-active" ""}}"
-            >{{ .name }}</button>
+            >
+                {{- if and (eq .type "default") (eq .fileIcon true) -}}
+                    <div class="tabs-nav__copy" data-tab-copy data-tab-copy-value="{{ .name }}">
```
JorindeUsMedia, line 25: a clickable item should be a button (accesiblity)., a div is not accessible by keyboard .

hugo/assets/ts/widgets/tabs.ts
```diff
@@ -56,6 +56,15 @@ export class Tabs extends BaseWidget {
                 }
             }
         });
+
+        // Listen to copybutton events
+        const copyButtons = this.element.querySelectorAll<HTMLElement>('[data-tab-copy]');
```
JorindeUsMedia, line 61: you could also use the copy widget by using data-copy instead of having seperate code for the tabs-copy. 
Would also be great if we can show a message when the item is copied. I think the copy widget already does that.

hugo/assets/scss/components/tabs-nav.scss
```diff
@@ -27,10 +28,27 @@
...
 
+    &__copy {
+        height: 36px;
+        position: relative;
+        width: 1rem;
+        z-index: 1;
+
+        &:focus,
+        &:hover {
+             pointer-events: cursor;
```
JorindeUsMedia, line 39: cursor is not a valid value. should be pointer?  better to use button instead of a div faking to be a button

JorindeUsMedia (changes requested):

content/examples/shortcodes/code-tabs/en.md
JuanUsMedia: Check!

hugo/assets/ts/widgets/tabs.ts
JuanUsMedia: Reverted this change and used the copy widget instead.

hugo/layouts/partials/tabs.html
JuanUsMedia: > updated to button for line 50-53 
> for line 25-29 a nested button inside another button is not valid or recommented. This causes rendingering and behavior issues. I replaced it with a span element.

JorindeUsMedia (approved):
Looks good!

Nice to have (non-blocking) : show a message when the filename is copied like we also do for the code copy:  
Screenshot from 2024-08-30 10-08-02

jpluscplusm:
@JuanUsMedia Once all conversations are resolved, I can review and import into Gerrit.

jpluscplusm (changes requested):
Hey @JuanUsMedia - thanks for your work on this!

I think it looks fab, and the end result for me is great - but only because I opened the initial issue and I know what I'm looking for on the page! I believe we need some additional UI elements to make this feature useful to the reader who *doesn't* know that it exists ...

1. A tooltip-on-hover needs to present some short indication that a copy-on-click is going to happen, or the user will have no idea that they can engage the feature without first accidentally clicking the file icon. Is displaying a tooltip when the mouse is over the relevant action area an option? Text would be short, e.g. "Copy filename", or "Click to copy filename".
1. Jorinde already mentioned the idea of a on-click "Copied!" indication. If this is visually and JavaScript-ly compatible with point #<wbr>1, then I feel it would be useful. If it's not, then the tooltip would be the more important indicator, to my mind.

Please do give me a nudge if you'd like to discuss this, and don't burrow down a JS/etc rabbit-hole if a chat might simplify the process!

JorindeUsMedia:
> Hey @JuanUsMedia - thanks for your work on this!
> 
> I think it looks fab, and the end result for me is great - but only because I opened the initial issue and I know what I'm looking for on the page! I believe we need some additional UI elements to make this feature useful to the reader who _doesn't_ know that it exists ...
> 
> 1. A tooltip-on-hover needs to present some short indication that a copy-on-click is going to happen, or the user will have no idea that they can engage the feature without first accidentally clicking the file icon. Is displaying a tooltip when the mouse is over the relevant action area an option? Text would be short, e.g. "Copy filename", or "Click to copy filename".
> 2. Jorinde already mentioned the idea of a on-click "Copied!" indication. If this is visually and JavaScript-ly compatible with point #1, then I feel it would be useful. If it's not, then the tooltip would be the more important indicator, to my mind.
> 
> Please do give me a nudge if you'd like to discuss this, and don't burrow down a JS/etc rabbit-hole if a chat might simplify the process!

@JuanUsMedia 
1. Can be done by adding a title attribute. I'm just not sure if that works on a non button or a element.. 
2. I think only needs an html & css addition. The js part for showing a message after copying is already there

JuanUsMedia:
Hi @jpluscplusm @JorindeUsMedia, 

1. A title attribute with the text "Copy filename" has been added.
2. A "Copied!" message appears when a user clicks on the file icon. The copy widget is used for the JavaScript part.

JorindeUsMedia (approved):

jpluscplusm (approved):
This looks great - thank you! I'll import the PR into Gerrit..

jpluscplusm (changes requested):
I'm getting some `npm run lint` errors in CI: https://github.com/cue-lang/cuelang.org-trybot/actions/runs/10671895886/job/29578446764.

They look fixable at the code level, but I'm definitely no expert in this area - please could you take a look @JuanUsMedia? Do let me know if they look odd/unexpected/etc, and we can hop on a call to discuss :-)

JuanUsMedia:
Hi @jpluscplusm, Thanks for pointing that out, I’ve fixed the linter issues!

PR #470, opened 2024-09-19
hugo: Fix linewrap codeblock

Wies-UsMedia:
- Adds white-space: pre on base code block element for proper linewrapping on Chrome.

To test: go to https://cuelang.org/docs/howto/embed-files-in-cue-evaluation/ and find the line with `type=<filetype>` in the Chrome browser. This should not wrap over 2 lines anymore.

Fixes [#3287](https://github.com/developer-usmedia/cuelang.org/compare/feature/3287)

### DIFF
--- a/hugo/assets/scss/base/base.scss
+++ b/hugo/assets/scss/base/base.scss
@@ -340,6 +340,7 @@ code {
     font-style: normal;
     font-weight: inherit;
     line-height: inherit;
+    white-space: pre;
 
     ul &,
     ol &,


hugo/assets/scss/base/base.scss
```diff
@@ -336,6 +336,7 @@ pre {
 }
 
 code {
+    display: inline-flex;
```
JorindeUsMedia, line 339: this is now breaking the usage of code in some other places like https://deploy-preview-470--cue.netlify.app/examples/shortcodes/code-tabs/ and https://deploy-preview-470--cue.netlify.app/docs/howto/transform-json-with-cue/  where the columns are now not growing correctly anymore and the code is now displayed in a row because of the inline-flex (when you turn of this css line you can see the difference on this pages) . Can't we use inline-block? or maybe not change the display style but use white-space: pre or nowrap? We need to make sure the solution doesn't break other usages of the code tag.  

Screenshot from 2024-09-19 15-03-46

JorindeUsMedia (changes requested):

hugo/assets/scss/base/base.scss
Wies-UsMedia: Replaced the display with white-space pre instead!

JorindeUsMedia (approved):

jpluscplusm (changes requested):
The change works perfectly, here on Chrome(OS).

Because GitHub doesn't give me a space to review your commit message, I've copied it here verbatim:

> - Adds white-space: pre on base code block element for proper linewrapping on Chrome.

Please wrap the body of commit messages after 80 characters.

> To test: go to https://cuelang.org/docs/howto/embed-files-in-cue-evaluation/ and find the line with "type=<filetype>" in the Chrome browser. This should not wrap over 2 lines anymore.

This link goes to the current (unfixed) cuelang.org, but we prefer to note the path in a form that's easily copy-and-paste-able onto a preview version of the site. For example:

```
To test, go to `/docs/howto/embed-files-in-cue-evaluation/` and find the line
with "type=<filetype>" in the Chrome browser. This should not wrap over 2 lines
anymore.
```

> Fixes https://github.com/developer-usmedia/cuelang.org/compare/feature/3287

This needs to be written as:

```
Fixes cue-lang/cue#3287
```

jpluscplusm (approved):
LGTM! 

I slightly misspoke: we wrap after character 78, not 80. I've fixed this up whilst importing the PR into Gerrit as https://cuelang.org/cl/1201785

PR #471, opened 2024-09-20
typo: fixed docs/modules Build Constraints external link

yevvhen:


jpluscplusm:
Thanks for pointing this out, @yevvhen. I'm not sure which location @rogpeppe intended this link to point to, so I'll ask them to confirm before we import this into Gerrit!

myitcv:
I believe this change has since been superseded. Thanks for flagging in the first place, @yevvhen

PR #474, opened 2024-10-25
hugo: Modified blue tint for base styling

Wies-UsMedia:
As per commit. 

To test and compare, go to: 
$PREVIEW/examples/shortcodes/info
$PREVIEW/examples/shortcodes/buttons


JorindeUsMedia (commented):
Question: is $c-blue still used anywhere after this change?
If not I would suggest assigning the value that is now assigned to $c-blue--light to $c-blue and getting rid of --light

Wies-UsMedia:
> Question: is $c-blue still used anywhere after this change? If not I would suggest assigning the value that is now assigned to $c-blue--light to $c-blue and getting rid of --light

Updated accordingly.

hugo/assets/scss/config/colors.scss
```diff
@@ -2,7 +2,7 @@ $c-blue--lightest:      #f8f9fd;
 $c-blue--lighter:       #f0f2fb;
 $c-grey--light-mid:     #e6e8f3;
 $c-blue--light:         #1b3987;
```
JorindeUsMedia, line 4: Now blue--light and blue are the same. Also the variable $c-grey--light-mid is wrongly named because it's also one of the blues. So let's clean this up a bit because otherwise it's confusing for the future:

$c-blue--light:         #1b3987;  is now a duplicate of $c-blue and can be removed.
$c-grey--light-mid > rename to $c-blue--light

So after that change we will have: 

```$c-blue--lightest:      #f8f9fd;
$c-blue--lighter:       #f0f2fb;
$c-blue--light:         #e6e8f3;
$c-blue:                #1b3987;
$c-blue--bright:        #063AC0;
$c-blue--dark:          #0c2c65;
$c-blue--darker:        #0f1333;
```

JorindeUsMedia (changes requested):

hugo/assets/scss/config/colors.scss
Wies-UsMedia: Updated colors file and one affected scss file (trees).

JorindeUsMedia (changes requested):
The $c-blue--light from before which is been replaced by $c-blue is not changed in all scss files yet. I now get the very light blue as nav text for instance. Please check all usage of $c-blue--light and check if it needs to be replaced with $c-blue now. 

image

image

Wies-UsMedia:
> The $c-blue--light from before which is been replaced by $c-blue is not changed in all scss files yet. I now get the very light blue as nav text for instance. Please check all usage of $c-blue--light and check if it needs to be replaced with $c-blue now.
> 
> image
> 
> image

I believe I caught all of them now, but let me know if you still find a few out of place.

JorindeUsMedia (approved):

PR #477, opened 2024-11-01
hugo: Allow preselected tags on search

Wies-UsMedia:
See commit message.

To test, go to https://deploy-preview-477--cue.netlify.app/examples/shortcodes/search/ and you should be able to add new tags, as well as remove the preselected tags.

### DIFF
--- a/hugo/assets/ts/widgets/search-filter.ts
+++ b/hugo/assets/ts/widgets/search-filter.ts
@@ -1,8 +1,8 @@
 // eslint-disable-next-line
 import Fuse from 'fuse.js';
-import { BaseWidget } from './base-widget';
-import { FilterEvent, FilterItem, ParsedQuery, SearchEvents } from '../interfaces/search';
 import { parseQuery, queryToUrlParams } from '../helpers/search';
+import { FilterEvent, FilterItem, ParsedQuery, SearchEvents } from '../interfaces/search';
+import { BaseWidget } from './base-widget';
 
 export class SearchFilter extends BaseWidget {
     public static readonly NAME = 'filter';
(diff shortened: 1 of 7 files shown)


JorindeUsMedia (approved):

hugo/assets/ts/widgets/search-filter.ts
```diff
@@ -139,21 +140,40 @@ export class SearchFilter extends BaseWidget {
...
 
-        if (!query) {
-            this.selectedItems = [];
-        } else {
-            this.selectedItems = this.parsedQuery.facets[this.filterName] ?? [];
+        let urlParamsChanged = false;
+        let selectedItems = this.parsedQuery.facets[this.filterName] ?? [];
+
+        // Get preselected items if no items set in url (if set in urL: url wins from preselected)
+        if (selectedItems.length === 0 && this.element.dataset.preselectedItems && this.element.dataset.preselectedItems !== '') {
```
JorindeUsMedia, line 164: As discussed in our call: we should only add preselected tags  when there is no query at all and not just when there is no tags in the query. So change it to if !(query) I guess or !this.parsedQuery ( not sure what the value is when there is no query)

JorindeUsMedia (changes requested):

hugo/assets/ts/widgets/search-filter.ts
Wies-UsMedia: Should be updated now.

jpluscplusm:
@Wies-UsMedia Is this ready for (re-)review?

Wies-UsMedia:
> @Wies-UsMedia Is this ready for (re-)review?

It is yep!

jpluscplusm:
CI passed. Preview is here: https://cl-1203609-1--cue-cls.netlify.app/

jpluscplusm (approved):
Reviewed via https://cl-1203609-1--cue-cls.netlify.app/examples/shortcodes/search/.

We still have a corner-corner case where removing the last tag leaves an empty `q` query string param, but then resets the tags to the predefined set, but that's sufficiently niche that I'll just track it as a new issue that we can address if it becomes a problem.

PR #479, opened 2024-11-08
Fix wrong bounds for uint16 and uint32

andrea-amadei-supercell:
Inconsistent according to CUE own definition [here](https://github.com/cue-lang/cue/blob/af459f14b88efc59bee88f3b3a0f8b1b37dbfa66/internal/core/export/bounds.go#L185).


jpluscplusm (approved):
Thanks, @andrea-amadei-supercell  - good catch!

I'll run this PR through our preprocessor, and import the result into Gerrit. The fix should go live sometime later today.

jpluscplusm:
Thanks again for this, @andrea-amadei-supercell. I've imported it as https://cuelang.org/cl/1203855.

I had to reset the author on your commit (back to me), as your DCO/`Signed-off-by`-line didn't match your committer email address (perhaps because you made the change via the GitHub webUI?) and our process requires that these 2 strings match precisely. I hope that's ok - and thanks for the fix!

PR #484, opened 2024-12-18
playground: Fix loading of an example while id is in the url.

JorindeUsMedia:
- Loading an example should replace the url: not add a parameter to it.
- If both id & example are in the url params: remove the example parameter from the url

To test:
Go to /play?id=eVdbhSP1QnG&example=test-example#w=function&i=cue&f=eval&o=cue > the example parameter should be removed Click on an example: the example should load and replace the saved code

Fixes: cue-lang/cue#2994


jpluscplusm (approved):
LGTM. Importing now.

jpluscplusm:
Imported as https://cuelang.org/cl/1206292.

PR #486, opened 2024-12-18
hugo: show background color on all inline code blocks

anne-usmedia:
* Removed the code that only adds a background to inline code when inside p or list
* Give code element a transparent background when inside of pre element.

Test:
* Add inline code to some steps and see if they have a background color:
/examples/shortcodes/step/
* Check all other uses for code, not only inline, e.g.:
    * /examples/basic/code-block/
    * /examples/shortcodes/code-tabs/

Fixes
https://github.com/cue-lang/cue/issues/3003
and
https://github.com/cue-lang/cue/issues/3498

### DIFF
--- a/hugo/assets/scss/base/base.scss
+++ b/hugo/assets/scss/base/base.scss
@@ -336,22 +336,16 @@ pre {
 }
 
 code {
+    --pre-background-color: #{ $c-grey-blue };
+
+    background-color: var(--pre-background-color);
+    border-radius: 2px;
     font-size: inherit;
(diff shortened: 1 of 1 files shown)


JorindeUsMedia (approved):

jpluscplusm (approved):
LGTM! Imported as https://cuelang.org/cl/1206742.

jpluscplusm:
After merging, I noticed a regression caused by this change. This is tracked as https://github.com/cue-lang/cue/issues/3656.

PR #487, opened 2025-01-07
hugo: add searchTerm option to search shortcode

JuanUsMedia:
* Add searchTerm to example page
* Pass searchTerm from search shortcode, through search partial, to searchbar partial
* Set initial state with searchItem

To test:
* Navigate to /examples/shortcodes/search and verify that the search results are accurate.

Fixes https://github.com/cue-lang/cue/issues/3566

### DIFF
--- a/hugo/assets/ts/widgets/search-autocomplete.ts
+++ b/hugo/assets/ts/widgets/search-autocomplete.ts
@@ -5,7 +5,7 @@ import { AutocompleteQuerySuggestionsHit } from '@algolia/autocomplete-plugin-qu
 import { BaseItem, GetSourcesParams, OnSubmitParams } from '@algolia/autocomplete-shared/dist/esm/core';
 import algoliasearch, { SearchClient } from 'algoliasearch';
 import merge from 'lodash.merge';
-import { getHiddenInputFacets, mapToAlgoliaFilters, parseQuery } from '../helpers/search';
+import { getHiddenInputFacets, mapToAlgoliaFilters, parseQuery, updateSearchParams } from '../helpers/search';
 import { SearchFacets, SearchItem } from '../interfaces/search';
 import { BaseWidget } from './base-widget';
 
@@ -21,6 +21,7 @@ export class SearchAutocomplete extends BaseWidget {
     private querySuggestionsPlugin: AutocompletePlugin<AutocompleteQuerySuggestionsHit, undefined>;
     private autocomplete: AutocompleteApi<BaseItem>;
     private readonly placeholder: string;
+    private readonly searchTerm: string;
     private readonly detachedModeMaxWidth = 1023;
     private readonly facetInputs: NodeListOf<HTMLInputElement>;
 
@@ -32,6 +33,7 @@ export class SearchAutocomplete extends BaseWidget {
         this.searchType = this.element.dataset.searchAutocomplete || '';
         this.searchbarSize = this.element.dataset.searchbarSize;
         this.placeholder = this.element.dataset.searchbarPlaceholder ?? '';
+        this.searchTerm = this.element.dataset.searchbarSearchTerm ?? '';
         this.facetInputs = document.querySelectorAll<HTMLInputElement>('[data-search-results] input[name^="facet-"]');
     }
 
@@ -55,6 +57,7 @@ export class SearchAutocomplete extends BaseWidget {
     public init(): void {
         this.initQuerySuggestions();
         this.initAutocompleteSearch(this.searchClient, this.searchType);
+        this.initSearchTerm();
 
         // Prefill autocomplete with query from url on search results page
         if (this.searchType === 'results') {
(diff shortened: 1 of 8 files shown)


jpluscplusm:
Hey @JuanUsMedia - I believe the CI failure is caused by you needing to run the preprocessor to align the contents of `content/` and `hugo/`.

JuanUsMedia:
Hi @jpluscplusm, 

Thanks for your message. When I attempt to run the preprocessor, I encounter the following error: 
`error loading .cue files: github.com/cue-lang/cuelang.org/internal/ci.github.com/cue-lang/cuelang.org/internal/ci/base: import failed: import failed: import failed: cannot find package "github.com/cue-tmp/jsonschema-pub/exp1/githubactions": cannot fetch github.com/cue-tmp/jsonschema-pub/exp1/githubactions@v0.3.0: module github.com/cue-tmp/jsonschema-pub/exp1/githubactions@v0.3.0: 401 Unauthorized: unauthorized: bearer auth required`

Do you have any idea where this issue might be coming from or how to solve it?

hugo/assets/ts/widgets/search-autocomplete.ts
```diff
@@ -238,6 +244,25 @@ export class SearchAutocomplete extends BaseWidget {
...
     }
 
+    private initSearchTerm() {
+        if (this.searchTerm) {
+            const url = new URL(window.location.href);
+            const searchParams = new URLSearchParams(url.search);
+            const query = searchParams.get('q');
+            const parsedQueryT = parseQuery(query);
+
+            this.storage.isAvailable = false;
```
JorindeUsMedia, line 254: why does this widget needs to set if storage is available? If you need sessionstorage instead of localstorage you can use sessionstorage directy without the helper. The storage helper is for localstorage and falls back on sessionstorage when localstorage is not available (incognito mode for instance). 

another question: why do we need storage for this instead of setting a boolean variable on the class itself?

hugo/assets/ts/widgets/search-autocomplete.ts
```diff
@@ -238,6 +244,25 @@ export class SearchAutocomplete extends BaseWidget {
...
+            const searchTermIsSet = this.storage.getItem(`searchterm-${this.searchTerm}`);
+
+            if (!searchTermIsSet) {
+                parsedQueryT.cleanQuery = this.searchTerm;
+                const updatedUrl = `${ window.location.href.split('?')[0] }?${ queryToUrlParams(parsedQueryT) }`;
```
JorindeUsMedia, line 259: nice to have / non-blocking:  move this to a helper function since search-filter.ts (updateUrl) also has something similar

JorindeUsMedia (changes requested):
I'm not getting any search results in the autocomplete and also no results in the search-results itself. I'm also thinking we're missing a piece of code for this in the search-results.ts?

hugo/assets/ts/widgets/search-autocomplete.ts
JuanUsMedia: Moved updateUrl logic to helper.

hugo/assets/ts/widgets/search-autocomplete.ts
JuanUsMedia: -  Sessionstorage is directly called instead of the storage helper.
- A boolean variable alone on the class will not prevent the search from 'looping'.

hugo/assets/ts/widgets/search-autocomplete.ts
```diff
@@ -238,6 +241,23 @@ export class SearchAutocomplete extends BaseWidget {
...
+            const url = new URL(window.location.href);
+            const searchParams = new URLSearchParams(url.search);
+            const query = searchParams.get('q');
+            const parsedQuery = parseQuery(query);
+
+            const searchTermIsSet = sessionStorage.getItem(`searchterm-${this.searchTerm}`);
+
+            if (!searchTermIsSet) {
+                parsedQuery.cleanQuery = this.searchTerm;
+                updateSearchParams(parsedQuery, false);
```
JorindeUsMedia, line 255: not sure if this is caused by this line but got a couple of bugs while testing:

To reproduce:
Go to https://deploy-preview-487--cue.netlify.app/examples/shortcodes/search in a new session (tab will do) 
Remove a tag: the tag gets removed but the intial query also get removed so after I remove the tag 'encoding' the search will update to just `tag:"cue command`.  
When I also remove that tag both tags get added again and I end up with ` tag:encodings tag:"cue command"`

When in the same session I go again to https://deploy-preview-487--cue.netlify.app/examples/shortcodes/search > the preset query is also gone.

I also think we're missing something in the search-results.ts.  It's a bit hard to do this I think because mulitple widgets (search-filters, search-autocomplete & search-results) now "fight"  over the params (tags, preselected tags, contenttype preselected hidden input etc) and we also repeat stuff in those widgets. Maybe we need to take a step back and check where is the best place to determine the (preselected) filters & query and only make 1 responsible for updating the url: Maybe a shared service?

PR #489, opened 2025-04-28
Hugo: add searchTerm option to search shortcode

JuanUsMedia:
* Add searchTerm to example page
* Pass searchTerm from search shortcode, through search partial, to filter partial
* Set initial searchItem with search filter

To test:
* Navigate to /examples/shortcodes/search and verify that the search results are accurate.

Fixes cue-lang/cue#3566


hugo/assets/ts/widgets/search-filter.ts
```diff
@@ -173,6 +173,11 @@ export class SearchFilter extends BaseWidget {
         this.selectedItems = selectedItems;
 
         if (urlParamsChanged) {
+            const initialSearchTerm = this.element.dataset.initialSearchTerm;
```
JorindeUsMedia, line 176: not sure if it's this line but when i only add searchTerm="json" to the shortcode and omit the preselected tags it doesn't search on 'json'. I think this should happen outside the if (urlParamsChanged)?

JuanUsMedia: You're right! I have updated the code bijmoving initialSearchTerm outside the if(urlParamsChanged)

JorindeUsMedia (approved):
The only edge case which I found while testing: Remove all preselected queries & tags: the search will reset to the preselected filters. There is no way to empty it. But I don't know how to fix that because on page load there is also no search term and than the preselected term kicks in: that's what we want. There's no way to know now if the user emptied it or if we started fresh.  It's also preselected for a reason :)

content/examples/shortcodes/search/en.md
```diff
@@ -24,3 +24,6 @@ showContentTypes
 
 tags
 : optional - Adds preselected tags
+
+searchTerm
+ : optional - Prefills the search bar
```
jpluscplusm, line 29: This doesn't render correctly - I think the leading space is the problem.

content/examples/shortcodes/search/en.md
```diff
@@ -8,10 +8,10 @@ With this shortcode you can embed a search results widget on the page.
 ## Example
 
 ```
-{{</* search contentType="How-to Guides" showContentTypes=false tags="encodings,commented cue" */>}}
+{{</* search contentType="How-to Guides" showContentTypes=false tags="encodings,commented cue" searchTerm="json" */>}}
```
jpluscplusm, line 11: Please change this example to have a multi-word `searchTerm` value instead of `json`. The search string `"built-in function"` would be compatible with the tags and content type. This is so that an author sees an example of how to encode a string that needs to be quoted **inside** a quoted Hugo parameter.

NB This isn't asking for the existing `json` string to be updated to `built-in function`. It's asking for an example of triggering Algolia's [advanced syntax](https://www.algolia.com/doc/api-reference/api-parameters/advancedSyntax/) to be present on the page, where the multi-word search string needs to be an exact match for a phrase on the page.

content/examples/shortcodes/search/en.md
```diff
@@ -8,10 +8,10 @@ With this shortcode you can embed a search results widget on the page.
...
 ```
 
-{{< search contentType="How-to Guides" showContentTypes=false tags="encodings,commented cue" >}}
+{{< search contentType="How-to Guides" showContentTypes=false tags="encodings,commented cue" searchTerm="json" >}}
```
jpluscplusm, line 14: Please update to match the updated example, above.

jpluscplusm (changes requested):

jpluscplusm (changes requested):
LGTM! Please just fix the DCO and my last comment, and I'll submit this :-)

content/examples/shortcodes/search/en.md
```diff
@@ -8,10 +8,10 @@ With this shortcode you can embed a search results widget on the page.
 ## Example
 
 ```
-{{</* search contentType="How-to Guides" showContentTypes=false tags="encodings,commented cue" */>}}
+{{</* search contentType="Tutorials" showContentTypes=false tags="encodings,cue command" searchTerm="\"json schema\" */>}}
```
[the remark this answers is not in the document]
JuanUsMedia: Added the missing double qoute and fixed the DCO :)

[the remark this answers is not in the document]
jpluscplusm: Thanks! The current CI failure is unrelated to your change, which looks good to me 👍

jpluscplusm (approved):
LGTM - I'll import into Gerrit after we've resolved the unrelated CI test failures.

PR #491, opened 2025-12-07
docs(installation): add mise alternative method installation

jylenhof:


jylenhof:
> @jylenhof Thanks for this! While you're here, could you please also document support for installation via [Scoop](https://scoop.sh)? It's basically the Homebrew for Windows and CUE is [available](https://github.com/ScoopInstaller/Main/blob/master/bucket/cue.json) in the main bucket. It's okay if you don't feel comfortable adding that in the same PR though!
> 
> ```
> ❯ scoop install cue
> Updating Scoop...
> Updating Buckets...
> INFO  Updating cache...
> Scoop was updated successfully!
> WARN  'cue' (0.15.4) is already installed.
> Use 'scoop update cue' to install a new version.
> ❯ scoop info cue
> 
> Name        : cue
> Description : CUE is an open source data constraint language which aims to simplify tasks involving defining and using data.
> Version     : 0.15.4
> Source      : main
> Website     : https://cuelang.org
> License     : Apache-2.0
> Updated at  : 1/27/2026 3:30:22 PM
> Updated by  : github-actions[bot]
> Installed   : 0.15.4
> Binaries    : cue.exe
> ```

I don't know scoop (not much a windows guy)... So better to create another PR I think

